PR #58, opened 2022-05-06
feat(Tag): init new component

### LINKED ISSUE #50: Add Tag Component

ghlost:
Start the skeleton for tag components

Resolves #50


johnpauldavis (changes requested):
@ghlost looking good, just a few minor tweaks and additions.

johnpauldavis:
@ghlost : 

1. To fix the build errors, run `pnpm run test:all` and commit what files get generated that are not gitignored.
2. Based on conversation with the ILO team in our last weekly, the Tag needs to detect if its url prop is populated. If not, it should output a span that displays the tag but without interactive states (i.e. no hover) and if it has a url, an anchor that does have the interactive states.

johnpauldavis (approved):

PR #71, opened 2022-05-26
fix(Accordion): spacing and color

ghlost:
Tweaked the spacing colors for accordion


ghlost:
@johnpauldavis I changed the name of base.json to _base.json because I was getting errors for referencing colors that weren't assigned (accordion.json comes before base.json so I couldn't reference base colors).

johnpauldavis (approved):

PR #77, opened 2022-06-17
fix(React) Breaking tests in Ubuntu

justintemps:


johnpauldavis (approved):

johnpauldavis:
@justintemps Just a note that since this is a branch off my Media PR, you may want to instead make a new branch off `develop` for just your change and merge _that_ in instead. That way my work on the Media component can still go through code review as per normal.

PR #78, opened 2022-06-23
feat(Callout): Callout component

ghlost:
Build out the twig and react callout component

Resolves Issue #76


packages/react/src/stories/Callout.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+  argTypes: {},
+  parameters: {
+    componentSubtitle: "Component",
+    docs: {
+      page: () => (
+        <>
+          <Subtitle />
+          <Title />
+          <Description>
+            The link component creates an anchor link with a specific theme.
```
johnpauldavis, line 42: This is still the Link description - maybe something like: The Callout component provides an inline alert or message to end users.

johnpauldavis (changes requested):

packages/react/src/stories/Callout.stories.tsx
ghlost: Fixed the copy

PR #100, opened 2022-07-28
Rebuild in netlify on each deploy

justintemps:
Rebuilds project in netlify on each deploy so that builds won't fail when you add commits to an open PR


justintemps:
So far, I've been able to recreate the problem by making a change (adding a console.log) to a .tsx file.

PR #108, opened 2022-08-04
Issue/ids 15/dropdown

johnpauldavis:
This adds the Dropdown component, a helper component called Fieldset (for wrapping form fields, their labels and all helper text in), and updates the Tooltip to allow an icon to be what it attaches to instead of a line of text.


packages/styles/scss/components/_dropdown.scss
```diff
@@ -0,0 +1,107 @@
...
+  background-image: url("#{colortodataurlicon("arrow", $color-ux-labels-actionable)}"),
+    linear-gradient(
+      to right,
+      transparent 0%,
+      transparent calc(50% - 0.81px),
+      $color-formelements-input-default-border-left calc(50% - 0.8px),
+      $color-formelements-input-default-border-left calc(50% + 0.8px),
+      transparent calc(50% + 0.81px),
+      transparent 100%
+    );
```
ghlost, line 44: Dang this is some tough math

ghlost (approved):

PR #109, opened 2022-08-04
bugfix(Accordion): fix tests

ghlost:
Remove testing for small accordions


ghlost:
Removed the small/large version per design feedback. It's now purely based on breakpoint.

johnpauldavis (approved):

PR #120, opened 2022-09-20
feat(Accordion): add variations to Twig Storybook for Accordion

johnpauldavis:
@ghlost Here's an example of variants with differing fields in the Accordion.


ghlost (approved):
this is great

PR #126, opened 2022-10-03
fix(Wingsuit): fix config issue

johnpauldavis:
This fixes an issue caused by naming conflicts in config files in Wingsuit


ghlost (approved):
Ahh this makes sense

PR #129, opened 2022-10-12
fix(Fonts):font wight ffix

johnpauldavis:
This fixes the font weight issue so that the font weights more closely match the Figma.


packages/fonts/index.scss
```diff
@@ -2,29 +2,19 @@
...
-  src: url("~@ilo-org/fonts/assets/Overpass-Regular.eot?#iefix")
-      format("embedded-opentype"),
-    url("~@ilo-org/fonts/assets/Overpass-Regular.woff2") format("woff2"),
-    url("~@ilo-org/fonts/assets/Overpass-Regular.woff") format("woff"),
-    url("~@ilo-org/fonts/assets/Overpass-Regular.ttf") format("truetype");
+  src: url("~@ilo-org/fonts/assets/Overpass-Medium.ttf") format("truetype");
   font-weight: normal;
   font-style: normal;
   font-display: swap;
 }
```
ghlost, line 13: Should this be set to font-weight 500 to match the design sizing?

PR #137, opened 2022-10-19
Feat/create font include for use outside monorepo

johnpauldavis:
This adjusts the Fonts and Styles packages to:

1. Include css for font inclusion that does not assume one is serving content from inside the monorepo
2. Output an alternative CSS file for that purpose
3. Offer a minified version of the CSS


justintemps:
Hey @johnpauldavis this PR needs a changeset before it can be approved.

justintemps (approved):

PR #149, opened 2022-10-27
Bugfix/jpd 10 27 22

johnpauldavis:
This includes bug fixes for Footer and Link List, as well as the addition of a mixin for applying the site wide transition effect asked for by design (see animations.scss for the mixin)


packages/twig/src/patterns/components/card/card.twig
```diff
@@ -1,6 +1,7 @@
 {#
   CARD COMPONENT
 #}
+<script>console.log({{ _context | json_encode | raw }});</script>
```
ghlost, line 4: Console log leftover

packages/twig/src/patterns/components/navigation/navigation.twig
```diff
@@ -1,6 +1,7 @@
 {#
   NAVIGATION COMPONENT
 #}
+<script>console.log({{ _context | json_encode | raw }});</script>
```
ghlost, line 4: same here

ghlost (changes requested):

PR #150, opened 2022-10-31
Bugfix/jm/10 26 22

ghlost:
Lots of bug fixes as well as some RTL specific styles for cards. 

- Added sizes for cards, certain cards have different padding or styles for wide/standard/narrow
- Changed the parameters for cardgroups, removing settings and adding type to denote variant of card
- Changed the way cornercut and dark settings work, including a :not selector for cards that do no utilize those features


johnpauldavis (approved):
Really nice work @ghlost - thank you

PR #164, opened 2022-11-03
Bugfix/jpd 11 03 22

johnpauldavis:
This fixes many bugs and adds right-to-left functionality for rich text, tabs, image and credit.


packages/icons/icons.yml
```diff
@@ -136,6 +136,12 @@
     - ILO Icons
   sizes:
     - glyph
+- name: listarrowreverse
+  friendly_name: listarrowsreverse
+  aliases:
+    - ILO Icons
+  sizes:
+    - glyph
```
ghlost, line 144: good call

packages/styles/scss/components/_richtext.scss
```diff
@@ -387,4 +387,58 @@
       }
     }
   }
+
+  .right-to-left & {
+    direction: rtl;
```
ghlost, line 392: nice

packages/twig/src/patterns/components/tabs/tabs.twig
```diff
@@ -6,7 +6,7 @@
 {% for item in items %}
   {% set tabids = tabids|merge([random(100000)]) %}
 {% endfor %}
-
+<div class="right-to-left">
```
ghlost, line 9: This looks like it was done for testing purposes, does this need to be removed?

ghlost (changes requested):

packages/twig/src/patterns/components/tabs/tabs.twig
johnpauldavis: This is to generate id attributes for the tabs since they're needed for aria.

But since we don't know how many tab components will be on a page, we can't assume that naming them after their labels or their index will be sufficiently unique.

ghlost (approved):

PR #177, opened 2022-11-10
Bugfix/jm/11 8 22

ghlost:


packages/twig/src/patterns/components/hero/hero.twig
```diff
@@ -1,15 +1,20 @@
...
-  {% include "@components/image/image.twig" with {
-    alt: image.alt,
-    caption: image.caption,
-    credit: image.credit,
-    prefix: prefix,
-    url: image.url,
-  } only %}
+
+  {% if not image %}
+    <figure class="{{prefix}}--bg" data-prefix="{{prefix}}">
```
johnpauldavis, line 7: Can you have this be a div when it's not holding an image? That way it will be semantically neutral.

johnpauldavis (changes requested):
looks good @ghlost just one minor tweak

packages/twig/src/patterns/components/hero/hero.twig
johnpauldavis: Even better!

johnpauldavis (approved):

PR #187, opened 2022-11-12
Build/upgrade pnpm@6.35.1

justintemps:
pnpm and changeset don't appear to be changing `workspace:*` to the latest version of the package in all cases. This problem is described [here](https://github.com/pnpm/pnpm/issues/4348) and the recommended fix is to upgrade pnpm. 

Here I'm recommending we upgrade to 6.35.1, which is the latest release for version 6, rather than version 7 with which we've encountered build issues in the past.


johnpauldavis:
@justintemps Is there a way for us to do a dry run of this on Netlify to make sure the build doesn't break?

justintemps:
> @justintemps Is there a way for us to do a dry run of this on Netlify to make sure the build doesn't break?

@johnpauldavis Yea, it's already done, that's what the Deploy previews are for.

johnpauldavis (approved):

Issue #270, opened 2023-01-20
React: add missing components


justintemps:
- [x] Navigation
- [x] Local Navigation
- [x] Breadcrumb
- [x] Card
- [x] CardGroup
- [x] Footer
- [x] Tabs

justintemps:
- [ ] Navigation
- [ ] Local Navigation
- [ ] Breadcrumb
- [ ] CardGroup
- [ ] Footer
- [ ] Tabs

justintemps:
@inesdgomes once you've added new categories, please specify where these components should go.

justintemps:
In Local Navigation, remove arrow from Main ILO Website Link. This needs to be done in `twig` as well

justintemps:
@wc-lei feel free to keep pushing up to the branch so we can review things bit by bit for now.

justintemps:
@wc-lei how about after the footer and localNav, we do the Cards next? The main navigation is really only for the main website which I don't see getting implemented in React any time soon.

justintemps:
hey @Shashika6, we can close this by merging #600. However, @bashlk had some feedback about the implementation that I think is right, we should adjust it as per his comments.

Shashika6:
Noted @justintemps

PR #290, opened 2023-02-10
React/feat/270

wc-lei:
Footer component implemented, please review.

Part of #270


packages/react/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,117 @@
...
+            src={logo}
+            alt="ILO Logo EN-White"
+          />
+          <h3 className={`${baseclass}--headline`}>{tagline}</h3>
+          <h4 className={`${baseclass}--subhead`}>{subtagline}</h4>
+        </div>
+        {address && (
+          <div className="address">
+            {address.map((line, i) => (
+              <p id={i.toString()} className="address--line">
```
justintemps, line 38: I think `id` should be `key`. Also keys can be numbers so no need to stringify. https://beta.reactjs.org/learn/rendering-lists#where-to-get-your-key

packages/react/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const baseclass = `${prefix}--footer`;
+
+  return (
+    <footer className={`${baseclass}`}>
+      <div className={`${baseclass}--main`}>
+        <div className="site--info">
+          <img
+            className={`${baseclass}--logo`}
+            src={logo}
+            alt="ILO Logo EN-White"
```
justintemps, line 30: May be just "ILO Logo". This will be read by screen readers so needs to be intelligible.

packages/react/src/components/Footer/Footer.args.ts
```diff
@@ -0,0 +1,75 @@
...
+          label: "Link Three",
+          url: "http://www.google.com",
+        },
+        {
+          label: "Link Four",
+          url: "http://www.google.com",
+        },
+        {
+          label: "Link Five Is Much Longer",
+          url: "http://www.google.com",
```
justintemps: Let's just https://www.ilo.org for all of the links

packages/react/src/components/Footer/Footer.args.ts
```diff
@@ -0,0 +1,75 @@
...
+        },
+      ],
+    },
+  ],
+  socials: [
+    {
+      label: "Facebook",
+      url: "https://www.facebook.com/",
```
justintemps: Use https://www.facebook.com/ILO.ORG/

packages/react/src/stories/Footer/Footer.mdx
```diff
@@ -0,0 +1,17 @@
+import { Meta, Story, Canvas, ArgsTable, Unstyled } from "@storybook/blocks";
+import * as FooterStories from "./Footer.stories.tsx";
+import { Footer } from "../../components";
+
+<Meta title="Footer" of={FooterStories} />
+
+# Footer Example
+
+The Footer component provides a Footer to users.
```
justintemps, line 9: Let's use the text from the Brand Hub:

```
The footer informs users about the Organization that a website belongs to. It can include copyright details, links to Terms and Conditions and generic calls to action. It should be included at the bottom of each page.
```

justintemps (changes requested):
Hey @wc-lei, this looks great, just a few minor tweaks to suggest.

justintemps:
@wc-lei also it looks like there's a rollup error to do with importing an svg, I can take a look at that.

packages/react/src/components/Footer/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as Footer } from "./Footer";
+export { default as FooterArgs } from "./Footer.args";
```
justintemps, line 2: Hey @wc-lei, this is the source of the rollup error. Rollup will bundle anything in `/components/**/index.ts`, so here it's trying to bundle the Args file which includes an SVG. But we don't want to bundle the Args files, those are only needed for Storybook and for Tests. So instead of exporting it here, just import it directly from the Storybook story or wherever else you need it.

justintemps:
@wc-lei, apologies, one more thing is that we need to add tests for this component.

justintemps:
Also and finally, we need to create a minor version bump for this. You can do that by running `npx changest` let me know if you want to do this together the first time.

wc-lei:
For LocalNav, I put in a very basic toggle handler but it doesn't seem to work, see if you can spot anything. Thanks.

packages/react/src/components/LocalNav/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as LocalNav } from "./LocalNav";
+export { default as LocalNavArgs } from "./LocalNav.args";
```
justintemps, line 2: Hey @wc-lei, just reviewing this now but I saw this pattern again in `LocalNav`. Remember that you shouldn't export the args files from component/whatever/index.ts or Rollup will try to bundle it.

packages/react/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,109 @@
...
+  subscribe,
+  legal,
+  secondarylinks,
+  anchorlink,
+}) => {
+  const { prefix } = useGlobalSettings();
+
+  const baseClass = `${prefix}--footer`;
+  const footerClasses = classNames(className, {
+    [baseClass]: true,
```
justintemps, line 25: I think classnames ignores falsey values so you could just write it like this:

```ts
const footerClasses = classnames(className, baseClass); 
```

packages/react/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,109 @@
+import { FC } from "react";
+import useGlobalSettings from "../../hooks/useGlobalSettings";
+import classNames from "classnames";
```
justintemps, line 3: As a convention I think we should import this as `classnames`

packages/react/src/components/LocalNav/index.ts
justintemps, line 2: @wc-lei This is probably why you're getting a rollup error. We should export the typs in `LocalNav.props.ts` but we should not export the test data in `LocalNav.args.ts`

justintemps (approved):

PR #309, opened 2023-03-02
Fix/twig: Unneeded props and markup in localnav

### LINKED ISSUE #294: Twig: LocalNav refactoring - remove irrelevant props and segments

justintemps:
Hey @wc-lei , I only saw a single section here that needed removing. Can you have a look?

Fixes #294


wc-lei (commented):
Do you see this segment? I don't think search is needed in localnav.

<div className={${prefix}--search}>
<button className={${prefix}--search--button} type="button">
{searchlabel}

Issue #299, opened 2023-02-24
Card: Content isn't visible on hover or is only visible on hover


justintemps:
A few examples in the Twig project. Good if we can tackle this together with #270. Adding @inesdgomes so she can check designs.

![Image](https://user-images.githubusercontent.com/12401179/221211197-81adc7df-f68d-4317-9394-fada9820f7a9.png)

![Image](https://user-images.githubusercontent.com/12401179/221211239-aefcc201-e144-4136-a1a5-7c9dd2a9e0fe.png)

![Image](https://user-images.githubusercontent.com/12401179/221211474-a272de12-f741-424b-a62b-731abbb31438.png)

wc-lei:
@justintemps @inesdgomes , I found this line of comment in the code: `The theme type for the card. Theme doesn't apply to these card types: Multilink, Data, Stat and Detail.`
The reason some texts are not shown in Multilink card is due to the font color (also in dark when the theme is dark). I have fixed that issue for GraphicText card, and introduced variation of font colours on either light or dark theme. 
It seems to me current implementation does not enforce a theme on card, is it true? Should we also vary font colour based on theme? In other words, allowing both light and dark theme for ALL cards.

justintemps:
This is fixed in #317

PR #317, opened 2023-03-27
feat(react): Implementation of Card and CardGroup components

wc-lei:
This commit also covers a patch fix issue #299.


justintemps:
@wc-lei we said this was just missing a tweak to the controls and could then be merged.

Issue #321, opened 2023-04-20
@storybook/addon-postcss@2.0.0 might be incompatible with Storybook 7


justintemps:
See [Storybook #20529](https://github.com/storybookjs/storybook/issues/20529)

```
Attention: We've detected that you're using the following addons in versions which are known to be incompatible with Storybook 7:

- @storybook/addon-postcss@2.0.0

Please be aware they might not work in Storybook 7. Reach out to their maintainers for updates and check the following Github issue for more information:
https://github.com/storybookjs/storybook/issues/20529
```

Integrayshaun:
Hey @justintemps 👋 

Storybook contributor here. While addon-postcss is deprecated, we plan to continue PostCSS through [`@storybook/addon-styling`](https://storybook.js.org/addons/@storybook/addon-styling)

PR #353, opened 2023-05-26
Use Vite builder with React Storybook and update pnpm

### LINKED ISSUE #326: Upgrade React package to Vite builder
This will improve build times and resolve annoying incompatibility between Webpack 5 in the React package and Webpack 4 in the Twig package.

justintemps:
The original scope of this PR was to hoist shared dependencies in the root package.json to make sure all the packages were using the same version of the most important ones.

However, this again led to conflicts between Wepback@4 in the Twig package and Webpack@5, which is a dependency of many of the packages we wanted to hoist. 

In an attempt to resolve this, I switched to building the `react` Storybook project with Vite, which alas, still won't enable us to hoist shared dependencies yet. We'll have to wait until the next version of Wingsuit for that.

However, it seemed worth doing in its own right and required several other fixes and adaptations which were all carried out.

**NB: This PR updates `pnpm` to version 8.6.0. You will have to update `pnpm` in your local environment to test this**. 

```bash
$ nvm use
$ npm i -g pnpm@8.6.0
```

This fixes #326 and #321


wc-lei:
@justintemps , this works fine for me, and the build seems faster. 👍

PR #355, opened 2023-05-30
React/feat/breadcrumb

wc-lei:
Hi @justintemps , can you have a look on this component? I think it's quite simple and straightforward but in the twig package there are some behaviour code that I don't understand, not sure if they are necessary.

### DIFF
--- a/.changeset/fresh-zebras-count.md
+++ b/.changeset/fresh-zebras-count.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/react": minor
+---
+
+feat(react): add breadcrumb as a new component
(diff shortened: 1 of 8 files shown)


justintemps (changes requested):
Hey @wc-lei, this looks good so far. The only thing it's missing is that >5 links, the breadcrumbs show an ellipses with a tooltip on click: 

Here's a live example: https://drupal-templates.ilo.org/templates/portal-with-nav

image

wc-lei:
@justintemps , implemented context menu behaviour for breadcrumb. Feel free to test or I can demo tomorrow.

justintemps:
Perfect @wc-lei thanks!

PR #360, opened 2023-05-31
Add block wrapping the hero image

PauZoocha:
Added a block to wrap the image on the Hero component. 
That allows Drupal to include responsive images.


justintemps:
Hey @PauZoocha, thanks for this! Awesome to see. Just a quick note that we try to use Angular style commit messages: https://gist.github.com/brianclements/841ea7bffdb01346392c

PR #377, opened 2023-06-20
feat(twig): Add if statement to all Drupal.behaviors

PauZoocha:
`Drupal.behaviors` does not only fire on page load, it fires on ajax calls or whenever new elements are inserted into the DOM. This adds an if statement checking if the js is proccessed to avoid calling the functions more than once, using data attributes.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -5,7 +5,7 @@
   "fixed": [],
   "linked": [],
   "access": "restricted",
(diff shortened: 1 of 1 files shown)


justintemps (changes requested):
@PauZoocha I don't think we should make changes to the .changeset config. Otherwise this looks ok.

justintemps (approved):

PR #386, opened 2023-06-22
feat(twig): Update cards image variable

PauZoocha:


justintemps:
This is handled by #388

PR #401, opened 2023-06-27
feat(twig): Add new Alt and Loading card variables into cardgroup template.

PauZoocha:
- Add **Alt** and **Loading** card variables into the card include on cardgroup template.


justintemps (changes requested):
This is great @PauZoocha thanks, could you also update `cardgroup.wingsuit.yml` so that these are actually passed to the component in Storybook and we make sure that it works?

justintemps (approved):

PR #403, opened 2023-06-27
feat(twig-styles): videojs and youtube.

PauZoocha:
- Change request for videojs player to be able to work with Youtube videos

### DIFF
--- a/packages/styles/scss/components/_video.scss
+++ b/packages/styles/scss/components/_video.scss
@@ -4,6 +4,8 @@
 
 .ilo--video {
   position: relative;
+  width: 100%;
+  max-width: 855px;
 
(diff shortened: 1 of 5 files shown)


packages/styles/scss/components/_video.scss
```diff
@@ -4,6 +4,8 @@
 
 .ilo--video {
   position: relative;
+  width: 100%;
+  max-width: 855px;
```
justintemps, line 8: @PauZoocha why is this needed? if you want the player to have max-width, just wrap it in a `div` no?

justintemps (changes requested):
Hey @PauZoocha just one thing for you to check here

PR #404, opened 2023-06-27
feat(twig): Change media attribute for responsive images

PauZoocha:
- Needed to change this attribute to be able to apply the breakpoint value that comes from Drupal - eg. `media= "all and (min-width: 480px)"`


justintemps:
Hey @PauZoocha can you explain the reasoning behind this? Is it because you might want to use `max-width` instead of `min-width`?

PR #441, opened 2023-07-28
fix(styles): Changed space between cards

rafaellapinho:
Added the same space between the cards, 32px.
Before:
image

After:
image


justintemps:
Thanks @rafaellapinho, long-term I think we should use tokens in the themes values to set constants like this, but given that the value was hard-coded before I think this is ok for now. cc @inesdgomes

PR #445, opened 2023-08-02
fix(styles): Removed max width of the cards

rafaellapinho:
Removed the max-width of the cards, so when we have only one card on the row, it can have the page container width.
Before: 
image

After:
image


justintemps:
This issue is fixed in #447

Issue #480, opened 2023-08-22
Add logo in additional languages to design system


inesdgomes:
Add the following versions of the ILO logo to the design system

- Arabic
- Chinese
- Dutch
- German
- Italian
- Japanese
- Portuguese
- Russian
- Turkish
- Ukrainian (TO BE CONFIRMED?)
- Vietnamese

All logos, apart from Ukrainian, are available in this folder:
[logos.zip](https://github.com/international-labour-organization/designsystem/files/12408631/logos.zip)

justintemps:
This was done in #506

PR #506, opened 2023-09-13
feat(brand-assets): #480 add additional logos in some languages

wc-lei:


justintemps:
@wc-lei this looks great, thanks so much for doing this. Before we merge, could you also update the table in packages/brand-assets/README.md?

justintemps (approved):

PR #600, opened 2023-10-09
feat(react/tabs) implemented tabs component in react

wc-lei:
Fixes :- https://github.com/international-labour-organization/designsystem/issues/270

A working first version of tabs component, improvements still needed.

- click on tab causing the tab component out of storybook
- so far, only richtext and image are recognized in tabs, if more components need to be embedded, the code needs to be updated

@Shashika6 
Picking up from where wei left - 

Fixes :- 

* Fixed clicking on tab causing the tab component out of storybook
* Fixed alignment with icon and label
* Passed the component directly instead of having two props as component and componentData
* Modified stories to match with twig
* Improvements to existing code


justintemps:
Hey @wc-lei is this ready for review or is it still a WIP?

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,103 @@
...
+            default:
+              return null; // Or handle unknown types as needed
+          }
+        })()}
+      </div>
+    </div>
+  );
+};
+
+// const Tabs: FC<TabsProps> = ({ items }) => {
```
bashlk, line 76: Commented out code.

packages/react/src/components/Tabs/Tabs.props.ts
```diff
@@ -0,0 +1,15 @@
...
+  /**
+   * Specify the items inside a tab
+   */
+  items: Required<Array<TabItem>>;
+}
+
+export interface TabItem {
+  icon?: IconProps;
+  label: Required<string>;
+  component: Required<string>;
```
bashlk, line 13: Why not pass in the component here directly? The component could then be placed directly in the content area.
content: <Text value="something" />

justintemps:
Hey @wc-lei what do you think of this feedback ☝️

justintemps:
Hey @Shashika6, awesome if you could pick this one up

Shashika6:
Will do @just intemps

packages/react/src/components/Tabs/Tabs.props.ts
Shashika6: @bashlk I think he has done this to keep it in sync with twig, I can do the change but that would require the change to be done in twig as well which could be a breaking change. What do you think ? Cc :- @justintemps

Shashika6: Also did the syncup to develop and made a few changes (Added them above)

bashlk: In this case, I think reworking the props to accept the component directly is better since it leads to a simpler implementation. It does mean that the props to the component will be slightly different than the twig version but here we are leveraging one of the features in React that makes it easier to achieve this kind of component composition. I am not sure whether we achieve the same pattern in twig in a nice way.

justintemps: Agree with @bashlk. In terms of keeping the two libs in sync, the React and Twig components don't necessarily have to take the exact same props. They do need to share the same stylesheets and we want to avoid styles specific to one implementation or the other. But we shouldn't feel like we have to implement the React components in a Twig way. React is React and when we can take advantage of it to simplify things, then we should, even if that makes the underlying logic different than the logic of the Twig component.

Shashika6: Got it guys will do the change thanks :)

Shashika6:
@justintemps This one is ready now made the changes based on feedback

PR #616, opened 2023-10-16
Fix(styles): Callout 

### LINKED ISSUE #549: Callout: various fixes
- [x] The Title alignment and the collapsible description are not matching with designs. The collapsible button is slightly up on the component.
- [x] Font family of the title doesn't match with the designs. It should be Overpass instead of Noto Sans
- [x] The symbol size on component is smaller than designs, not matching with screens and the page.

See screenshots below.

justintemps:
Fixes #549

@inesdgomes please check the deploy preview here: https://deploy-preview-616--ilo-ui-twig.netlify.app/?path=/docs/components-feedback-callout--callout

### DIFF
--- a/.changeset/quiet-horses-agree.md
+++ b/.changeset/quiet-horses-agree.md
@@ -1,0 +1,6 @@
+---
+"@ilo-org/styles": patch
+"@ilo-org/twig": patch
+---
+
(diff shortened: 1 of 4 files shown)


PauZoocha:
Looking good @justintemps

PR #627, opened 2023-11-07
fix(styles): fixed input border radius on iOS Safari

### LINKED ISSUE #546: Search bar looks incorrect in iOS
Corners look rounded for some reason

firefox_ZyzDxKKWlL
firefox_VD26jBHU1S

bashlk:
Fixes #546 

Upon investigation, the issue affected more than just the search field outlined in the issue and also affected all inputs and text areas. As such, this fix addresses all of it.

It is highly suspect that the CSS reset being used in this project does not already cover this. It seems like it is out of date. @justintemps  perhaps we should consider moving to a more maintained css reset.

### DIFF
--- a/.changeset/witty-trees-greet.md
+++ b/.changeset/witty-trees-greet.md
@@ -1,0 +1,7 @@
+---
+"@ilo-org/styles": patch
+"@ilo-org/twig": patch
+"@ilo-org/react": patch
+---
+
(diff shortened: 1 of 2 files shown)


justintemps (approved):
Hey @bashlk, this looks good. I opened a new issue (#628) to update the css reset.

Issue #571, opened 2023-09-28
Hero: In mobile, remove top spacing when no image


justintemps:
When there's no image in the Hero, we should remove the space for the image in mobile design. Or rather, the space should be less. See designs.

![image](https://github.com/international-labour-organization/designsystem/assets/12401179/44ad1a09-80dd-4d3f-bfe2-fea535087bc8)

**How it should be:**

![Image](https://github.com/international-labour-organization/designsystem/assets/22772102/8a3a66b0-f28e-465f-b047-f84abf83a54f)

inesdgomes:
@justintemps @Shashika6 With this fix, the space was completely removed but there should still be a gap of 48px between breadcrumb and hero on tablet and mobile. Does this need to be fixed in the design system or in the website?

**Website**
<img width="335" alt="Screenshot 2023-12-11 at 10 35 19" src="https://github.com/international-labour-organization/designsystem/assets/22772102/6cb6d9e8-013f-404f-b0f8-4f83fcc74d75">

**Design**
<img width="513" alt="Screenshot 2023-12-11 at 10 34 59" src="https://github.com/international-labour-organization/designsystem/assets/22772102/2675f4e1-3ff6-418e-8fad-2e558b50793a">

Shashika6:
@justintemps The breadcrumb gets hidden when the device is not large is this the expected behaviour ?

PR #636, opened 2023-11-09
Fix(Style): Remove top space in hero without image

Shashika6:
Problem

* Currently the grid row template assumes that an image always exists in css. Whether or not an image exists it will allocate a space for an image

Solution

* Added a separate class to handle styling when image does not exist

Development

* Modified css by creating a mixin which considers if an image exists or not by the class name. 

Screenshot 

Screenshot 2023-11-09 at 13 16 06

Issue :- https://github.com/international-labour-organization/designsystem/issues/571


packages/react/src/components/Hero/Hero.tsx
```diff
@@ -18,7 +18,7 @@ const HeroImage: FC<ImageProps> = ({ url, alt }) => {
   const defaultSrc = sortedUrls[sortedUrls.length - 1].src;
 
   return (
-    <figure className="hero--figure">
+    <figure className="hero with-image hero--figure">
```
bashlk, line 21: "with-image" is not really BEM compliant, I think you can manage with just "hero--figure".

Shashika6: @bashlk hero--figure currently exists for both if there is an image or if there isnt which is why i had to add a separate class.

packages/styles/scss/_mixins.scss
```diff
@@ -403,3 +403,11 @@
     }
   }
 }
+
+@mixin hero-grid-rows($max-height: $spacing-hero-image-max-height) {
+  grid-template-rows:
+    minmax($max-height, 1fr)
```
bashlk, line 409: Isn't this actually min-height? 😄

Shashika6: I was thinking the same xD, I didnt want to change the naming conventions they have already done so i was sticking to it but i'll refactor it

Shashika6: They have a variable defined as $spacing-hero-image-max-height

packages/react/src/components/Hero/Hero.tsx
bashlk: Had a quick call to discuss BEM and that we should probably refactor a bit or introduce a new modifier here.

packages/styles/scss/_mixins.scss
bashlk: If the existing names can be improved, we should go for it. 😄

in this case, if $spacing-hero-image-max-height is always referring to a min-height, then go ahead and rename it and then this param.

packages/react/src/components/Hero/Hero.tsx
Shashika6: Made the change :)

packages/styles/scss/_mixins.scss
Shashika6: Made the refactor of the variable

packages/react/src/components/Hero/Hero.tsx
```diff
@@ -18,7 +18,7 @@ const HeroImage: FC<ImageProps> = ({ url, alt }) => {
   const defaultSrc = sortedUrls[sortedUrls.length - 1].src;
 
   return (
-    <figure className="hero--figure">
+    <figure className="hero__row--with-image hero--figure">
```
bashlk, line 21: An element should not have multiple BEM element annotations, in this case, it is both hero__row and hero--figure. You can either add row to another element or add hero--figure__no-image for example.

Also pay attention to the separators, this project uses a modified BEM convention as noted in the outline - https://docs.google.com/document/d/18p5Z_d2gPE3gNWaVW5Ti6N65j8RR8Y0BqAdKEQt-yjk/edit?pli=1#heading=h.go781menmhq0

packages/styles/scss/components/_hero.scss
```diff
@@ -30,19 +30,25 @@
...
   overflow: hidden;
 
+  &__row {
+    &--with-image {
+      @include hero-grid-rows();
+    }
+  }
+
+  &:not(.hero__row--with-image) {
+    @include hero-grid-rows(1rem);
```
bashlk, line 43: Can't this be applied by default, to be overridden by the with image state?

packages/react/src/components/Hero/Hero.tsx
bashlk: Overall, I still think it would be better if we can separate the concerns and render a custom background component instead of the image component when no image is present.

Shashika6: oh got it now @bashlk will do the changes :)

Shashika6:
@bashlk made the changes to follow the BEM and made a small improvement to the component as well.

packages/react/src/components/Hero/Hero.tsx
```diff
@@ -58,7 +58,9 @@ const Hero: FC<HeroProps> = ({
 
   return (
     <div className={heroClasses}>
-      {image ? <HeroImage {...image} /> : <div className="hero--figure" />}
+      <div className="hero--figure">
+        {image ? <HeroImage {...image} /> : null}
```
bashlk, line 62: You can do a simpler image && <HeroImage /> here

packages/react/src/components/Hero/Hero.tsx
```diff
@@ -58,7 +58,9 @@ const Hero: FC<HeroProps> = ({
 
   return (
     <div className={heroClasses}>
-      {image ? <HeroImage {...image} /> : <div className="hero--figure" />}
+      <div className="hero--figure">
```
bashlk, line 61: hero--background would be a better name

packages/styles/scss/components/_hero.scss
```diff
@@ -30,19 +30,21 @@
...
-    minmax($spacing-hero-image-max-height, 1fr)
-    $spacing-hero-tooltip-height
-    $spacing-hero-card-corner-cut-height-sm
-    auto;
   grid-template-columns: 1fr;
   overflow: hidden;
+  @include hero-grid-rows(1rem);
+
+  &--image {
+    display: grid;
```
bashlk, line 38: Does the image itself really need this grid?

Shashika6: It does, Without it the styling gets off.

packages/react/src/components/Hero/Hero.tsx
Shashika6: Doing an && would return 0 so as it returns true or false so i think ternary operator would be the best in this case

packages/react/src/components/Hero/Hero.tsx
Shashika6: Sure

Shashika6: Will do, in that case will have to make the change in twig as well.

packages/react/src/components/Hero/Hero.tsx
Shashika6: Made the change :)

bashlk: Only if image is a number and its zero 😁

.changeset/spicy-beds-fix.md
```diff
@@ -0,0 +1,8 @@
+---
+"@ilo-org/react": minor
```
bashlk, line 2: React should probably also be a patch

bashlk (approved):

justintemps:
Hey @Shashika6 and @bashlk, thanks very much for your work on this. I'm working on another ticket related to the hero right now (#598). It's very tricky because the content in the Hero Card has to align with the content in the body of the website that's currently being developed and which uses the Design System. I need to merge that first before I merge this, so that will probably happen next week.

justintemps:
Hey @Shashika6 I was just testing this and it seems like it removes the top space even when there is an image.

image

Shashika6:
Figured out the issue @justintemps its structured a bit differently in react and twig ill put a quick fix for this now :)

Shashika6:
Can you check this out now @justintemps :)

justintemps:
Thanks @Shashika6 but now when I remove the image, I'm seeing the same problem as before...

image

Shashika6:
Made the change @justintemps can you verify now :)

justintemps:
Looks good @Shashika6, thanks this was a tricky one.

Issue #524, opened 2023-09-19
Re-implement RTL in styles


justintemps:
Right now, RTL styles are implement by adding a `.right-to-left` class to the root element. This is not the right way to do it. The idiomatic way of implementing an RTL design is by adding the [`dir` global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir).

This does a ton of the work out of the box and right now the `.right-to-left` class is fighting against it. 

## Components implicated

- Link List
- Navigation
- Language selector
- Hero Cards
- Highlighted card
- Footer

## More details

Example, here's the Footer as a child with the `.right-to-left` class. *This is the way it should look*

<img width="1683" alt="image" src="https://github.com/international-labour-organization/designsystem/assets/12401179/d5ae64cf-5cd2-4986-ab48-ea81d2419be3">

And here it is with the attribute and the class:

<img width="1685" alt="image" src="https://github.com/international-labour-organization/designsystem/assets/12401179/5fcb381b-126c-429a-9fb0-c51834274178">

And here it is with just the attribute:

<img width="1443" alt="image" src="https://github.com/international-labour-organization/designsystem/assets/12401179/035b8a6a-8f1c-4314-9a49-223ed4b486dd">

## Recommendation

We should go component by component, stripping out all use of `.right-to-left` and styling everything that needs RTL styles with the using the `dir` global attribute.

inesdgomes:
@justintemps Could we action this ticket in the coming days? It should be in the MVP

bashlk:
@justintemps can we start working on this or is there more details that need to be resolved?

Shashika6:
As discussed with @bashlk will be having a look at the entire codebase for the .right-to-left class and replacing it with the dir. Will investigate in the code base and add the other components below.

Shashika6:
The following components are using the right to left class. 

* Breadcrumb
* Card groups
* Credits
* Image
* Search field
* Text area
* Video

bashlk:
@justintemps I believe there are more components left that need to refactored to use dir.

Shashika6:
Yes @justintemps we thought of tackling individual components in separate PR's. Theres more components I have mentioned above that have not been mentioned as well.

justintemps:
Hey @Shashika6 just fyi, I'll handle the RTL styles for the Breadcrumb component.

Shashika6:
[...]

PR #651, opened 2023-11-10
Fix(Style): Support rtl in statcard

Shashika6:


packages/styles/scss/components/_statcard.scss
```diff
@@ -17,6 +17,10 @@
 
       @include cornercut(72px, 40px);
 
+      .right-to-left & {
```
bashlk, line 20: @Shashika6, take a look at #524 - we should be switching to using the dir attribute instead of using the .right-to-left class.

So let's use this as our first trial component, can you include this fix and also convert this component to use the dir attribute?

Shashika6: Will do @bashlk 🙌

Shashika6:
@justintemps I used the [dir='rtl'] attribute selector as @bashlk suggested I might need your help on testing this out.

justintemps:
Hey @Shashika6 and @bashlk, let's merge this while we're working on #524

packages/styles/scss/components/_statcard.scss
Shashika6: Going forward remove any css styles that are not needed.

bashlk (approved):

bashlk:
@justintemps I think we can merge this. Since we are going one component at a time with getting rid of .right-to-left, I think it would best to merge it as we go, rather than leaving all of these PRs open.

justintemps:
@bashlk and @Shashika6, it looks to me like the cornercut is still on the right of the card instead of the left?

image

packages/styles/scss/components/_statcard.scss
```diff
@@ -17,6 +17,10 @@
 
       @include cornercut(72px, 40px);
 
+      [dir="rtl"] & {
+        @include cornercut(72px, 40px, right);
```
justintemps, line 21: Pretty sure this should be 
```scss
@include cornercut(72px, 40px, left);
```

justintemps (changes requested):

packages/styles/scss/components/_statcard.scss
Shashika6: Sorry about this @justintemps will make the change :)

justintemps (approved):

Issue #521, opened 2023-09-19
Video: fix hover behaviour of play button


inesdgomes:
When we hover the duration of the video, the triangle should also change colour. The behaviour should be the same as when we hover the triangle.

https://github.com/international-labour-organization/designsystem/assets/22772102/0895bac9-3fa3-4765-85c1-e83e1c05ed7f

PR #653, opened 2023-11-13
fix(style): video timecode hover

GGKapanadze:
## Breakdown 
Let's break down this PR into 2 pieces, for `react` and `twig`. This separation is necessary because video components have drastically different implementations.

## Twig 
Twig component uses `video.js` for display, which has its own makeup for controls.
image
When we hover the button, duration is also styled because we are using the `sibling`(`~`) selector 
```scss
&:hover,
&:focus {
  background-color: $color-ux-video-controls-hover-background;
  @include iconbutton(
    "play",
    $spacing-ux-video-bigplaybutton-height,
    $spacing-ux-video-bigplaybutton-width,
    $color-ux-video-controls-hover-icon
  );

  & ~ .vjs-duration,
  & ~ .ilo--video--controls--duration {
    background-color: $color-ux-video-controls-hover-background;
    color: $color-ux-video-controls-hover-icon;
  }
```
but when we want to invert the behavior we can't use the `sibling` selector because the button is on top of `duration`  and sibling selectors can't select items above it
> The subsequent-sibling combinator (~) separates two selectors and matches all instances of the second element that follow the first element (not necessarily immediately) and share the same parent element.

Because of that we have Javascript implementation that listens to [mouseover](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseover_event) and [mouseout](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseout_event) on duration element and attaches `hovered` class to the button.

### Demo
https://github.com/international-labour-organization/designsystem/assets/36326203/48eb65b1-f4dd-470a-a3c4-13b1e06421f0

For implementation details, the `Video` class has the function `cacheDomReferences` but the current implementation is using `cacheVideoReferences` because of this chain 
```js
init() {
  this.cacheDomReferences()
    .start()
    .cacheVideoReferences()
    .setupHandlers()
    .enable();

  return this;
}
```
We can't find dom references for `video.js` until it's instantiated.

### a11y 
Since button element is first and it's tab indexed there is no problem for `keyboard` a11y

https://github.com/international-labour-organization/designsystem/assets/36326203/80f05f67-d728-4afd-905c-c7a7740874a2

----
## React
The current react implementation has no implementation for this feature at all, so I included that one too. 

### Current
https://github.com/international-labour-organization/designsystem/assets/36326203/1ed2dd8f-1319-4f7e-a87c-0addc5625b87

### Fixed
https://github.com/international-labour-organization/designsystem/assets/36326203/60fa0434-78f0-43f4-9ff6-94fa42a93a5f

but on the react side everything is upside down compared to Twig, I mean that controls are reversed 
```jsx
<>
  <label
    className={`${controlsClasses}--duration ${showposter ? "show" : ""}`}
  >
    {duration}
  </label>
  <button
    className={`${controlsClasses}--${!playing ? "play" : "pause"}`}
    onClick={togglePlay}
  >
    <span>
      {!playing ? controls && controls.play : controls && controls.pause}
    </span>
  </button>
</>
```
so without breaking an old implementation, the initial flow was done by a sibling selector but reversed, and the problematic part was done with the same behavior as the Twig one. 

For implementation details, I'm handling `mouseover` and `mouseout` directly on the button without controlling the hover effect with `state`, because there is no need to bring extra rerenders for the `hover` effect, since the application logic does not need it

### a11y
Since the button element is the second one we also handling `onFocus` and `onBlur` for keyboard navigation 

https://github.com/international-labour-organization/designsystem/assets/36326203/18f4885d-bf54-4e6c-b2d8-2fbed58f8758

Closing #521

### DIFF
--- a/.changeset/metal-peas-exercise.md
+++ b/.changeset/metal-peas-exercise.md
@@ -1,0 +1,7 @@
+---
+"@ilo-org/react": minor
+"@ilo-org/twig": minor
+"@ilo-org/styles": patch
+---
+
+Added chained hover effect for video play button
--- a/packages/react/src/components/Video/VideoPlayer.tsx
+++ b/packages/react/src/components/Video/VideoPlayer.tsx
@@ -1,4 +1,11 @@
-import { FC, createRef, useState } from "react";
+import {
+  FC,
+  FocusEvent,
+  MouseEvent,
+  createRef,
+  useCallback,
+  useState,
+} from "react";
 /* temporary way of importing ReactPlayer due to a known issue with ReactPlayer.
  * Revert to standard method of importing once RP's dev has fixed.
  */
@@ -166,6 +173,27 @@ const VideoPlayer: FC<VideoPlayerProps> = ({
     setSeeking(false);
   };
 
+  const handlePlayHover = useCallback(
+    (
+      event: MouseEvent<HTMLButtonElement> | FocusEvent<HTMLButtonElement>,
+      state: boolean
+    ) => {
+      const element = event.currentTarget;
+      if (!element.classList.contains(`${controlsClasses}--play`)) {
+        return;
+      }
+
+      const duration = element.previousSibling;
+      if (duration instanceof HTMLLabelElement) {
+        duration.classList.toggle(
+          `${controlsClasses}--duration--hovered`,
+          state
+        );
+      }
+    },
+    [controlsClasses]
+  );
+
   return (
     <div
       className={`${baseClass}--container`}
@@ -214,6 +242,10 @@ const VideoPlayer: FC<VideoPlayerProps> = ({
         <button
           className={`${controlsClasses}--${!playing ? "play" : "pause"}`}
           onClick={togglePlay}
+          onMouseOver={(e) => handlePlayHover(e, true)}
+          onFocus={(e) => handlePlayHover(e, true)}
+          onMouseOut={(e) => handlePlayHover(e, false)}
+          onBlur={(e) => handlePlayHover(e, false)}
         >
           <span>
             {!playing ? controls && controls.play : controls && controls.pause}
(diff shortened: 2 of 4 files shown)


bashlk:
Great work on this @GGKapanadze , I see that it was really tricky given the different implementations across twig and react. I don't like that we have to resort to JS to mend this, it adds a lot of complexity to the component. 

Just an idea, but what if we add a custom hover class to each of these elements (e.g. .hover-group) and then handle the hover on the parent element to apply styles to this hover class? (e.g. .parent:hover > .hover-group)

GGKapanadze:
Before the JS solution, I tried the grouping strategy as you mentioned, but there were sizing problems.

## React 
- On a smaller screen the video wrapper is spanned at full width so hovering will behave unexpectedly

https://github.com/international-labour-organization/designsystem/assets/36326203/89fee314-6e3e-4c11-a700-fee477338dd9

But it can be fixed by either reworking the markup a bit or managing the width with CSS

## Twig 
- Default size wrapper is higher than necessary elements

https://github.com/international-labour-organization/designsystem/assets/36326203/0ed18f47-22c2-4e5d-8d5f-c84c08fa397b

- The Responsive version behaves the es same way as the react

https://github.com/international-labour-organization/designsystem/assets/36326203/7fd38ff1-e1cb-47c5-b8f9-aaa190e69c82

These issues are fixable without JS but it will be a bit clanky because the wrapper contains not only duration and play a button and we have to size them properly, so I decided to go with a bit future-proof solution but I agree about complexity concerns.
image

Do you want me to go with the CSS approach and then discuss which one fits our needs @bashlk ?

justintemps:
Hey @bashlk, I think @GGKapanadze's solution is fine. Like the description says, the Video component is one of just a few that have completely different JS implementations in React and Twig, as long as they're still using the same CSS classes I'm ok with it.

PR #656, opened 2023-11-14
docs(twig/react): implemented addon to switch text direction

### LINKED ISSUE #655: Storybook: Add an addon for switching the text direction between ltr and rtl
Most of the components in the design system should also work in rtl layouts so it would be useful to be able to easily toggle between ltr and rtl layouts within storybook itself. This can be done by adding an addon for storybook to both the react and twig storybooks (e.g. https://storybook.js.org/addons/storybook-addon-rtl-direction)

bashlk:
Added storybook-addon-rtl-direction to both the twig and react storybooks. Now we can switch between ltr and rtl within storybook itself. The addon changes the dir attribute on the html element in the iframe that displays the content of the storybook.

https://github.com/international-labour-organization/designsystem/assets/11556301/41190489-40d6-4708-828d-a4897a1ac42b

Fixes #655

### DIFF
--- a/packages/react/.storybook/main.ts
+++ b/packages/react/.storybook/main.ts
@@ -27,6 +27,7 @@ export const addons = [
   "@storybook/addon-essentials",
   "@storybook/addon-a11y",
   "@storybook/addon-styling",
+  "storybook-addon-rtl-direction",
 ];
 
 export const typescript = {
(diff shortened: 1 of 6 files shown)


justintemps:
@bashlk very cool!

justintemps:
@bashlk also for future reference, the scope for commits that only affect storybook should be `docs(react)` or `docs(twig)` because the Storybook projects are the documentation for the packages.

PR #657, opened 2023-11-14
Fix(Style): Refactor RTL to use dir in herocard

Shashika6:

Development


bashlk (approved):

bashlk:
Just a quick note @Shashika6 that it will be better if you reference the issue with a # instead of the link, that way the reference will be tracked in the issue itself. Also it would be better if you can post a screenshot of the component in LTR and RTL.

PR #658, opened 2023-11-14
Fix(Style): Refactor RTL to use dir in image

Shashika6:

Development


bashlk:
@Shashika6 Since this component is broken in rtl without the changes to the credit, I think it would be better if you close that PR and move the changes in it to this one.

bashlk (approved):

PR #659, opened 2023-11-14
Fix(Style): Refactor RTL to use dir in credit

Shashika6:

Development


Shashika6:
This change will be done in the following [merge request](https://github.com/international-labour-organization/designsystem/pull/658)

Issue #639, opened 2023-11-09
Revise components spacing: Feedback category


inesdgomes:
Revise all spacings in the **Feedback components' category**:

- [ ] Callout
- [ ] Modal
- [ ] Tooltip
- [ ] Notifications

All spacings were slightly updated to follow the follow the 8 or 4px rule. See local variables in [Foundations file](https://www.figma.com/file/qr8jQnxqJkRJFQnVC5PmrB/ILO-Foundations?type=design&node-id=34-1554&mode=design&t=MqBqsc05HIxPHUo2-0):

<img width="1258" alt="spacing" src="https://github.com/international-labour-organization/designsystem/assets/22772102/7a4262fc-a41c-455c-9768-baed29e403ba">

Issue #637, opened 2023-11-09
Revise components spacing: Cards category


inesdgomes:
Revise all spacings in the **Cards components' category**:
- [ ] Card
- [ ] Card group
- [ ] Data card
- [ ] Detail card
- [ ] Fact list card
- [ ] Feature card
- [ ] Multilink card
- [ ] Promo card
- [ ] Stat card
- [ ] Text card

All spacings were slightly updated to follow the follow the 8 or 4px rule. See local variables in [Foundations file](https://www.figma.com/file/qr8jQnxqJkRJFQnVC5PmrB/ILO-Foundations?type=design&node-id=34-1554&mode=design&t=MqBqsc05HIxPHUo2-0):

![Image](https://github.com/international-labour-organization/designsystem/assets/22772102/a3ca91a0-1373-4ecb-b545-35fa13dd2fc6)

bashlk:
@inesdgomes Note that we can't see the variables in the foundation file as viewers. I requested edit access but I think I would be better if there was an explicit sizes table on the Spacing page.

GGKapanadze:
@inesdgomes I have marked points for refactoring but need access to the foundation file you mentioned.

justintemps:
Hey @inesdgomes, I've given @bashlk, @GGKapanadze and @Shashika6 edit access to the [Foundations File](https://www.figma.com/file/qr8jQnxqJkRJFQnVC5PmrB/ILO-Foundations?type=design&node-id=1485-257&mode=design&t=84IQCofOp2yy3DkH-0) so they can see the variables. 

However, I think @bashlk is right, we should add a sizes table on the Spacing Page. I've created a new issue for this, let me know if you want to discuss: https://github.com/international-labour-organization/designsystem/issues/675

beatrizmartinmartins:
Hey everyone @justintemps @bashlk @GGKapanadze @Shashika6 I have added the Spacing table on this link [https://www.figma.com/file/qr8jQnxqJkRJFQnVC5PmrB/ILO-Foundations?type=design&node-id=1567%3A249&mode=design&t=Pdy4nZQPLZ9FOs2T-1](https://www.figma.com/file/qr8jQnxqJkRJFQnVC5PmrB/ILO-Foundations?type=design&node-id=1567%3A249&mode=design&t=Pdy4nZQPLZ9FOs2T-1) CC @inesdgomes

GGKapanadze:
[...]

PR #660, opened 2023-11-15
refactor(themes): added new config values to theme

GGKapanadze:
## Description

This PR is a gateway for upcoming global refactoring of spacing 
- https://github.com/international-labour-organization/designsystem/issues/637
- https://github.com/international-labour-organization/designsystem/issues/643
- https://github.com/international-labour-organization/designsystem/issues/642
- https://github.com/international-labour-organization/designsystem/issues/641
- https://github.com/international-labour-organization/designsystem/issues/640
- https://github.com/international-labour-organization/designsystem/issues/639
- https://github.com/international-labour-organization/designsystem/issues/638

This will need follow-up PR that will do dead token elimination


justintemps:
Hey @GGKapanadze I think we're going to need all of the potential spacing values between spacing-1 and spacing-28. 

Also, note that we're going to have to delay killing off the dead tokens as it seems like they're being used by the new website. This would represent a breaking change, so we'll have to do this carefully. cc @bashlk

bashlk:
@justintemps Then perhaps we should make a separate issue for getting rid of the dead tokens and document the dead tokens there.

PR #680, opened 2023-11-17
Fix(Style): Support rtl in dropdown

Shashika6:
Issue :- [#524](https://github.com/international-labour-organization/designsystem/issues/524)

Development

* Modified css to use dir attribute selector and change the side of the dropdown based on direction

* LTR

Screenshot 2023-11-18 at 03 16 36

* RTL

Screenshot 2023-11-18 at 03 17 06

### DIFF
--- a/.changeset/happy-cheetahs-marry.md
+++ b/.changeset/happy-cheetahs-marry.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Refactor RTL style using dir in dropdown
--- a/packages/styles/scss/components/_dropdown.scss
+++ b/packages/styles/scss/components/_dropdown.scss
@@ -130,4 +130,18 @@
     );
     @include bordervalues("input", "formelements", "error");
   }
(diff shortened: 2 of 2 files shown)


Shashika6:
@bashlk @justintemps  I did the base implementation of the RTL and it looks good as the screenshots above.

However I noticed that the hover and active states are using the border set by passing in the formcontrols(tokens) which has css like border-left: 4px. 

LTR
Screenshot 2023-11-18 at 03 46 02

RTL

Screenshot 2023-11-18 at 03 46 12

How should i proceed with this should I create new form controls for RTL or is this okay for the time being ?

bashlk:
@Shashika6 This is a question for @inesdgomes. IMO, I think we should change the form control states so that the border is on the right on RTL.

I think you can proceed with merging this. If we decide to change the form control states, we can do it within a different issue where it can be changed for all the components (since many form controls have this coloured border)

Shashika6:
Noted @bashlk

PR #685, opened 2023-11-19
Fix(Style): Refactor rtl to use dir in video

Shashika6:
Issue :- [#524](https://github.com/international-labour-organization/designsystem/issues/524)

Development

* Convert right to left class to use dir selector attribute
* Fix styles to align play button and duration to the right when rtl is set. (Small/Large screen)

LTR (Small screen)

Screenshot 2023-11-20 at 00 42 35

RTL (Small screen)

Screenshot 2023-11-20 at 00 42 43

LTR (Large screen)

Screenshot 2023-11-20 at 00 35 34

RTL (Large screen)

Screenshot 2023-11-20 at 00 35 45

### DIFF
--- a/.changeset/smart-flies-thank.md
+++ b/.changeset/smart-flies-thank.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Refactor rtl to use dir in video
--- a/packages/styles/scss/components/_video.scss
+++ b/packages/styles/scss/components/_video.scss
@@ -119,6 +119,10 @@
       }
     }
 
+    [dir="rtl"] & {
+      justify-content: flex-end;
+    }
+
     button span {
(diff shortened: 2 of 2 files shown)


justintemps:
Question @inesdgomes, do we need to invert the direction of the video controls in RTL layouts? It looks like @Shashika6 has moved the initial play button to the right side of the video player, but the controls still have the play/pause button on the left. Does that get switched around as well or can we leave it as is?

inesdgomes:
Hi @justintemps Did some quick research and it seems videos in RTL should **not** be mirrored. See section 'The passage of time': _Anything depicting the passage of time should be mirrored. Do not mirror media playback buttons and the media progress indicator as they refer to the direction of tape being played, not the direction of time._ https://m2.material.io/design/usability/bidirectionality.html#mirroring-elements

PR #700, opened 2023-12-03
Ended video play button size 

GGKapanadze:
# Ended video play button size 

Fix the play button size for the ended video.

### Issue

Luckily we don't have a Drupal CSS re-bundled issue 🤞.

The problem is that the storybook is running in dev mode and the ended video still has paused class 

image

But the production bundle has a `playing` class on the ended video
image

And the style targeted the `paused` state 
image

I can investigate more into `video.js` implementation if we want to solve it on the library level @justintemps 

### Demo 

> Demo is running on local Drupal install 

https://github.com/international-labour-organization/designsystem/assets/36326203/78cea777-c2b5-4f7b-bfc2-5ae0139034c3

### DIFF
--- a/.changeset/rare-points-clean.md
+++ b/.changeset/rare-points-clean.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+reduced play button size for ended video
--- a/packages/styles/scss/components/_video.scss
+++ b/packages/styles/scss/components/_video.scss
@@ -186,7 +186,8 @@
         }
       }
 
-      &.vjs-paused {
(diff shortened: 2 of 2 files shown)


bashlk:
@GGKapanadze Do the CSS classes behave differently when storybook is running in dev mode?

GGKapanadze:
@bashlk if you mean `.vjs-ended`, the answer is yes it's consistent. As I mentioned the `video.js` has inconsistent toggle of `.vjs-playing` and `.vjs-paused` on ended video and the styling was done by `.vjs-paused`

Issue #532, opened 2023-09-21
React/Twig: Fix tooltip styles


justintemps:
In both libraries:

- [x] the position of the icon is wrong
- [x] the arrow is pointing in the wrong place
- [x] the box needs to be variable according to the amount of content there is
- [ ] the tooltip only shows when we hover the 'i' symbol, but should show when we hover any part of the dark grey shape
- [x] there is a weird behaviour on hover - see video below

<img width="502" alt="image" src="https://github.com/international-labour-organization/designsystem/assets/12401179/4c79f271-3f9c-4c27-aabe-9d00b0a70290">

https://github.com/international-labour-organization/designsystem/assets/22772102/e55226bf-473e-4775-a81e-ce9df4241931

bashlk:
@justintemps @inesdgomes Can you elaborate on what you mean when you say "the box needs to be variable according to the amount of content there is"? 

Does this mean that the tooltip should not have a max width and should expand in a single line to fill the entire window size and then wrap when it goes beyond it? 

Or is the tooltip not intended to be used with very long text and what is the typical length expected to be used in it?

justintemps:
@GGKapanadze in standup suggests using standard scale values to handle the tooltip size (`sm`, `lg`, etc..) that would determine width may be also padding. Then the actual position of the tooltip is determined automatically based on how much space is around it.

For the sake of simplicity, let's start with 2 variants. 

cc @inesdgomes

bashlk:
@Shashika6 I think you can start on this. We can address the following points without input from design.
 the position of the icon is wrong
 the arrow is pointing in the wrong place
 the tooltip only shows when we hover the 'i' symbol, but should show when we hover any part of the dark grey shape
 there is a weird behaviour on hover - see video below
 
For now assume that there is a 100px and 130px width variant of the tooltip as indicated in figma but be ready to change these widths later. 

Also assume that tooltip should be displayed on the right of the hover target by default and then top or bottom if there is no space on the right. I noticed that there is no design for the top and bottom direction for the 130px variant so assume that it is the same as the 100px version but taller.

inesdgomes:
Hi @bashlk 

1. We're going to update the tooltips' design documentation to reflect the various possible widths and in which cases they should be used, but we won't be able to action it this week. While we don't have detailed guidance, could we make the size change according to the content with two thresholds:
- If text is shorter than 50 characters, max. width is 180px
- If text is longer than 50 characters, max. width should be 400px
[...]

PR #701, opened 2023-12-03
fix(styles): React/Twig : Tooltip 

Shashika6:
Fixes [#532](https://github.com/international-labour-organization/designsystem/issues/532)

### To do :- 

- [x] the position of the icon is wrong
- [x] the arrow is pointing in the wrong place
- [x] the box needs to be variable according to the amount of content there is
- [x] there is a weird behaviour on hover 

### Notes 

* Implemented [Popper.js](https://popper.js.org/) to handle the tooltip.
* Modified styling and js as required to support popper.js

Cc :- @justintemps

### Screenshots

Long tooltip (More than 50 chars)

Screenshot 2023-12-03 at 21 37 54

Normal tooltip (Less than 50 chars)

Screenshot 2023-12-03 at 21 38 31

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react/src/components/Tooltip/Tooltip.tsx
@@ -21,6 +21,7 @@ const Tooltip: FC<TooltipProps> = ({
   });
   const [arrowPlacement, setArrowPlacement] = useState<string>("center");
   const [arrowAlignment, setArrowAlignment] = useState<string>("left");
+  const [tooltipSize, setTooltipSize] = useState<string>("");
 
   const tooltipRef = useRef(null);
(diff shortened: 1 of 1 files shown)


Shashika6:
The grey area mentioned is a completely separate component (Caption). The tooltip component is just the icon. 

Screenshot 2023-12-04 at 21 57 30

So when we hover over the grey area it doesn't trigger the tooltip. In react we could use something like a forward ref to achieve the same hover effect in the child component in the parent component as well but I don't think its quite possible in twig. 

Cc :- @justintemps @bashlk

justintemps:
hey @Shashika6 looks like this is still a WIP right?

Issue #573, opened 2023-09-28
Card hover animations seem out of sync


justintemps:
Many different cards, the hover animations just don't look right.

bashlk:
@justintemps It is difficult to understand what you mean by smoother here, can you quantify it somehow?

justintemps:
List of affected cards:
- Feature Card
- Detail Card

PR #719, opened 2023-12-08
fix(styles)- Make card animations smoother

Shashika6:
Fixes :- https://github.com/international-labour-organization/designsystem/issues/573

### Notes 

* Made transition-timing smoother
* Fixed light theme hover issue

### DIFF
--- a/.changeset/hot-cherries-grow.md
+++ b/.changeset/hot-cherries-grow.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Make card animations smoother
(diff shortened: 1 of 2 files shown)


Shashika6:
@justintemps Can you verify if this fixes the issue. I increased the transition timing to make the effect more smoother

justintemps:
Hey @Shashika6 this does indeed fix the issue but it looks like we're missing transition animations on:

- Promo Card
- Text Card

Could we add those too?

Btw @inesdgomes Should we add a filter to the image in the Multilink Card on hover like we do with Detail Card and Feature Card?

https://deploy-preview-719--ilo-ui-twig.netlify.app/?path=/docs/components-cards-multilink-card--multilink-card

Shashika6:
Added transition animates on Promo card and text card as well @justintemps

PR #721, opened 2023-12-12
Updated CSS Reset

GGKapanadze:
## Updated CSS Reset

The new implementation consists of two parts 

- The first part is to normalize elements across browsers using [normalize.css](https://github.com/necolas/normalize.css) which is a somewhat standard part of modern CSS resets
- The second part has **opinionated** reset that will create a good enough base reset for `designsystem` without any major breaking changes, so for that reason, I decided to go with a custom implementation, partially utilizing [destyle.css](https://github.com/nicolas-cusan/destyle.css)

## Checks

- I tried to fast-check every component, and no major issues were found, but we still need one good cross-browser check for a second opinion 
- Additional ios+safari demo from the native device for #546 

https://github.com/international-labour-organization/designsystem/assets/36326203/5813a521-2a94-4471-9810-a3b56b29c32e


justintemps:
This will require extensive, manual design review from @inesdgomes to check for regressions.

justintemps:
@GGKapanadze to rebase so that we can test this in the ILO website.

justintemps:
Hey @GGKapanadze, we talked on Teams. I tried to review this, but for some reason I'm not able to run the Drupal site with BB8, I get a database connection error. Could you see if it's happening to you as well? I'm getting the same error on my local installation, so think they've probably made some configuration change.

Issue #534, opened 2023-09-21
React: Update Tooltip styles to sync with Twig


justintemps:
See more info here: https://github.com/international-labour-organization/designsystem/pull/533

Shashika6:
@justintemps When you mean by sync does it mean just the icon to be fixed or the entire popper tooltip functionality to be included as well.

justintemps:
With popper as well. The implementation should be as similar as possible.

Shashika6:
Noted @justintemps

PR #737, opened 2023-12-18
fix(react) - Sync tooltip with twig

Shashika6:
Fixes :- https://github.com/international-labour-organization/designsystem/issues/534

### Notes 

* Implemented [Popper.js](https://popper.js.org/) to handle the tooltip.
* Modified classes to fix icon

### Screenshots

Long tooltip (More than 50 chars)

Screenshot 2023-12-18 at 23 36 28

Normal tooltip (Less than 50 chars)

Screenshot 2023-12-18 at 23 29 48

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.args.ts
+++ b/packages/react/src/components/Tooltip/Tooltip.args.ts
@@ -22,7 +22,7 @@ const dark: TooltipProps = {
 const darkicon: TooltipProps = {
   icon: true,
   label: "This is a tooltip for an icon",
-  theme: "dark", // "light" | "dark" ;
+  iconTheme: "dark", // "light" | "dark" ;
   isVisible: false,
 };
 
--- a/packages/react/src/components/Tooltip/Tooltip.props.ts
+++ b/packages/react/src/components/Tooltip/Tooltip.props.ts
@@ -1,4 +1,4 @@
-import { TooltipThemes } from "../../types";
+import { TooltipThemes, TooltipIconThemes } from "../../types";
 
 export interface TooltipProps {
   /**
@@ -34,5 +34,5 @@ export interface TooltipProps {
   /**
    * Describe the background theme of the tooltip
    */
-  iconTheme?: Required<TooltipThemes>;
+  iconTheme?: Required<TooltipIconThemes>;
 }
--- a/packages/react/src/components/Tooltip/Tooltip.props.ts
+++ b/packages/react/src/components/Tooltip/Tooltip.props.ts
@@ -27,22 +27,12 @@ export interface TooltipProps {
   isVisible?: boolean;
 
   /**
-   * Callback fired onMouseOver
-   */
-  onMouseOver?: (e: React.MouseEventHandler<HTMLDivElement>) => any;
-
-  /**
-   * Callback fired onMouseOut
-   */
-  onMouseOut?: (e: React.MouseEventHandler<HTMLDivElement>) => any;
-
-  /**
    * Describe the theme of the tooltip
    */
   theme?: Required<TooltipThemes>;
 
   /**
-   * Set a width for the tooltip
+   * Describe the background theme of the tooltip
    */
-  width?: string;
+  iconTheme?: Required<TooltipThemes>;
 }
--- a/packages/react/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react/src/components/Tooltip/Tooltip.tsx
@@ -1,138 +1,119 @@
-import { FC, useState, useRef } from "react";
+import React, { FC, useState, useRef } from "react";
 import classNames from "classnames";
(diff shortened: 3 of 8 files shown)


Shashika6:
@justintemps In react it has the tooltip for both text and icon. Should I just leave the label as it is or just have the icon like the twig component. Also @justintemps might need your help in making the width bigger for the story in tooltip for react.

packages/react/src/components/Tooltip/Tooltip.args.ts
```diff
@@ -22,7 +22,7 @@ const dark: TooltipProps = {
 const darkicon: TooltipProps = {
   icon: true,
   label: "This is a tooltip for an icon",
-  theme: "dark", // "light" | "dark" ;
+  iconTheme: "dark", // "light" | "dark" ;
```
justintemps, line 25: @Shashika6 why are we changing this to `iconTheme`?

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,138 +1,119 @@
...
   const baseClass = `${prefix}--tooltip`;
   const [isVisible, setIsVisible] = useState<boolean>(false);
-  const [position, setPosition] = useState<{ x: number; y: number }>({
-    x: 0,
-    y: 0,
-  });
-  const [arrowPlacement, setArrowPlacement] = useState<string>("center");
-  const [arrowAlignment, setArrowAlignment] = useState<string>("left");
+  const [longTooltip, setLongToolTip] = useState<boolean>(false);
+  const [popperInstance, setPopperInstance] = useState<PopperInstance | null>(
```
justintemps, line 20: @Shashika6 & @bashlk are we sure we want to do this with state? Would it not be better to use a ref to store the instance?

justintemps:
> @justintemps In react it has the tooltip for both text and icon. Should I just leave the option to have a tooltip in a label as it is or just have the icon only like the twig component.
> Screenshot 2023-12-19 at 00 26 19
> 
> Also @justintemps might need your help in making the width bigger for the story in tooltip for react.

Sure let's talk during the standup today

justintemps (changes requested):

packages/react/src/components/Tooltip/Tooltip.args.ts
Shashika6: The theme is for the tooltip. I added a new prop called iconTheme which is available in twig as well. Since this is dark icon I added the iconTheme.

packages/react/src/components/Tooltip/Tooltip.tsx
bashlk: Yeah this would be better as a ref since the popper instance is mutable ref.

packages/react/src/components/Tooltip/Tooltip.args.ts
bashlk: So the twig tooltip has support for applying a different theme to the icon and then the tooltip itself?

packages/react/src/components/Tooltip/Tooltip.tsx
Shashika6: @bashlk I noticed that we can just use a function to check on the fly. Something like 

```suggestion
  const isLongTooltip = () => {
    const tooltipText = (tooltipRef.current?.textContent || tooltipRef.current?.innerText || '').trim();
    return tooltipText.length > 50;
  };
```

What do you think ?

packages/react/src/components/Tooltip/Tooltip.args.ts
justintemps: Oh yea that's right. The icon can be a different color than the tooltip.

packages/react/src/components/Tooltip/Tooltip.tsx
bashlk: Yeah that could be alright - my only concern is that the ref could be undefined while this is evaluates by I guess this runs onMouseOver when the ref has already been mounted.

Shashika6: Yup it runs on mouseOver so in that case will be removing the state

packages/react/src/components/Tooltip/Tooltip.props.ts
```diff
@@ -26,23 +26,13 @@ export interface TooltipProps {
...
    * Describe the theme of the tooltip
    */
   theme?: Required<TooltipThemes>;
 
   /**
-   * Set a width for the tooltip
+   * Describe the background theme of the tooltip
    */
-  width?: string;
+  iconTheme?: Required<TooltipThemes>;
```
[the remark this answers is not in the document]
Shashika6: Sure will have its own set of types

justintemps (approved):

PR #750, opened 2023-12-21
feat(twig): Update Hero component Media queries Breakpoints

PauZoocha:
- When applying the breakpoints on the Hero component, the twig template markup forces to always apply `(min-width: "")` rule and from Drupal the media queries come out depending the devices sometime applying `(max-width: "") for mobile.

Before: 
`(min-width: all and (max-width: 414px)px)` 

After: 
` all and (max-width: 414px)`


justintemps:
Hey @PauZoocha, let's talk about this. We're a little worried that adding this will lead to lots of disparate media-queries getting used that don't have anything to do with the standard breakpoints provided by the design system. Can you provide some more context about what you're trying to accomplish?

justintemps:
Thanks @PauZoocha, your latest commit seems like a good compromise. @bashlk fyi see https://github.com/international-labour-organization/designsystem/pull/750/commits/e077aa74060f0433e2648f3ba0a387be02c9cd97

bashlk:
I think this will still lead to breakpoint widths being scattered across the site codebase, leading to the hero component behaving differently than documented in the design, as both the site code and design evolve over time.

@PauZoocha Is drupal changing the media queries of the source element somehow? Or are you refering to some other styles that you have on the site that interferes with this?

justintemps:
@PauZoocha correct me if I'm wrong, but I think the Drupal module wants to pass in a media query rather than an integer. 

@bashlk I forgot that this is already implemented in this way in the [Picture component](https://github.com/international-labour-organization/designsystem/blob/develop/packages/twig/src/patterns/components/picture/picture.twig), so @PauZoocha is just following the technique we've already used elsewhere. 

I can see why you're concerned about a website's code base using a bunch of breakpoints that don't match the DS's but it also seems like 1) we don't want to be fighting against Drupal and 2) this only related to responsive images and so doesn't affect the layout of components.

With this solution we're letting people do two things:

- Pass in an integer that we'll use to set a min-width (already implemented)
- Pass in a breakpoint that will set the responsive image based on arbitrary parameters (@PauZoocha's addition, but already implemented in the Picture)

May be a solution could be to _also_ let the user pass in a standard breakpoint setting. These are set in the themes package like this:

```json
{
  "breakpoints": {
    "x-small": { "value": "320px" },
    "small": { "value": "414px" },
    "medium": { "value": "610px" },
    "large": { "value": "1024px" },
    "x-large": { "value": "1168px" },
    "xx-large": { "value": "1440px" }
  }
}
```

So we'd let users do any of the following:

- Pass in the keyword of a breakpoint (`xsmall`, `small`, `medium`, `large`, `xlarge`, `xxlarge`) to set a min-width based on the DS breakpoints
- Pass in an integer to set min-width
- Pass in an arbitrary media query

The easiest option becomes to use the site breakpoints, but you can also use different ones if you need to for some reason. Thoughts @bashlk ?

bashlk:
@justintemps I think it is getting too complicated now. 

Since this only applies to the breakpoints of the responsive images, the proposed solution could be OK. 

I am still vary of having values like this outside of the design system though.

Issue #742, opened 2023-12-19
Styles: Feature Card links moves on hover in RTL layout 


justintemps:
Some legacy CSS changes the padding on the text in the CTA at the bottom of the card so that when you hover on it, the link moves. This only happens in RTL mode.

![image](https://github.com/international-labour-organization/designsystem/assets/12401179/72643a00-2cb0-4ff4-8ff6-1bf0b484a2e0)

Issue #716, opened 2023-12-05
Stat card: line moves when we hover the hyperlink


inesdgomes:
When we hover the hyperlink at the bottom of the card, the purple bottom line of the card moves down.

https://github.com/international-labour-organization/designsystem/assets/22772102/3ad9d7a1-225f-4ee1-9870-7b4f0c74ee18

PR #751, opened 2023-12-22
Fix(style) - Hover in stat card link

Shashika6:
Fixes :- https://github.com/international-labour-organization/designsystem/issues/716
Fixes :- https://github.com/international-labour-organization/designsystem/issues/742

Notes :- 

* Currently on hovering the link the card moves down a bit due to the border size changing. 

Recording :- 

https://github.com/international-labour-organization/designsystem/assets/32934169/4cbb6cfc-6a65-4fe7-878c-ec9c32638808

### DIFF
--- a/packages/styles/scss/_mixins.scss
+++ b/packages/styles/scss/_mixins.scss
@@ -231,7 +231,7 @@
 
   &:hover {
     background: map-get($color, "link", "background", "hover");
-    border-bottom: $widths-border-med solid
+    border-bottom: $widths-border-sm solid
       map-get($color, "link", "underline", "hover");
     color: map-get($color, "link", "text", "hover");
     text-decoration: none;


Shashika6:
@justintemps I noticed that when we click on the link the card goes down. Should we fix that as well or is that the desired state. 


https://github.com/international-labour-organization/designsystem/assets/32934169/67426ba2-494f-40b7-a837-5182a1697607

Shashika6:
Changed it to use text-decoration-offset and text-decoration-thickness @justintemps :)

justintemps:
Thanks @Shashika6, it looks like the line is supposed to be blue and then when you click it, the whole thing (including the underline has a yellow background)

image

Shashika6:
Added a small spacing at the bottom so that the underline is also included in the background @justintemps

justintemps:
@Shashika6 Found another problem

image

Think we need to make the line height a little bigger, about 24px or so I think.

justintemps:
One more thing here, @Shashika6 this change doesn't just affect the statcart component, it also affects links so we need to make it clear what the scope of this work is in the changeset

Shashika6:
Screenshot 2024-01-09 at 21 06 56

@justintemps This is how it looks with line height of 24px. Also updated the changeset.

justintemps:
Perfect! Thanks @Shashika6

PR #752, opened 2023-12-22
fix(style) - Feature card RTL link 

Shashika6:
Fixes :- https://github.com/international-labour-organization/designsystem/issues/716

Notes :- 

* Currently the feature card does not have padding assigned to link in RTL. This needs to be refactored.

Fix :- 

* Add appropriate padding for RTL 

Recording :- 

https://github.com/international-labour-organization/designsystem/assets/32934169/02cda9d7-b600-4c16-b9fb-015f6766ea09

### DIFF
--- a/.changeset/thirty-steaks-melt.md
+++ b/.changeset/thirty-steaks-melt.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Fix link position on RTL on feature card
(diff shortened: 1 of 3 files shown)


justintemps:
Hey @Shashika6, let's use padding-inline-start and padding-inline-end to handle these situations:

https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-start

Shashika6:
Refactored to use padding-inline to fix the above issue @justintemps

Issue #470, opened 2023-08-21
React: Refactor Cards to match new Twig implementation


justintemps:
See #466, #467, #454, and especially #447

PR #756, opened 2024-01-07
fix(react) - Sync cards with twig

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/470

### Notes :- 

* Currently the cards look broken and very out of sync with twig
* The profile in the text card will be fixed in https://github.com/international-labour-organization/designsystem/pull/754

### Checklist :- 

- [x] Move cards to individual folders
- [x] Remove unused props and values in each individual card
- [x] Refactor stories to remove the default card and have cards for each story
- [x] Compare with twig if implementation is properly in sync

### Fix :- 

* Refactored names to match with twig.
* Refactored layout synced up and added new classes from twig.
* Refactored card args. 
* Added storybook block for cards.

### Screenshots :- 

Screenshot 2024-01-07 at 19 36 12


Shashika6:
@justintemps I noticed that the link component is also somewhat out of sync with twig. (Link component is available in the data card)

React :- 
Screenshot 2024-01-07 at 19 49 35

Twig :- 
Screenshot 2024-01-07 at 19 50 00

justintemps:
Hey @Shashika6 thanks for your hard work on this. I think the implementation needed to follow the Twig version a little more closely, right now we've got a very complicated card component whereas I think it would be more useful to have a different component for each kind of card. Then, like with the Twig version, you could either use them separately or via a main <Card /> component that would render the one you wanted and pass it all the arguments (via a Switch statement or something).

Shashika6:
Makes sense @justintemps will have a look at moving the cards to separate components.

Shashika6:
@justintemps Did the initial implementation to separate the cards will need to do a bit of testing and also address the following issue :- https://github.com/international-labour-organization/designsystem/issues/637. Let me know what you think about this approach ?

Shashika6:
@justintemps Can you have a look at this one changes are pushed

Shashika6:
@justintemps This is ready to be reviewed. :)

justintemps:
Hey @Shashika6 , I don't think so it looks like in the website the pseudoelement is being rendered outside of the component. Can you try to figure out why?

image

justintemps:
@Shashika6  We agreed we need to use `type` here instead of `variant` for the main card component to be consistent with the Twig implementation.

Shashika6:
Change is done @justintemps renamed the variant prop to type to be consistent with twig.

bashlk:
@justintemps I took a quick look at the card components and I think the combined card component which allows a card type to be specified is a big reason why the mess arose originally. Do we really need it, now that we will have the individual cards available separately?

I would propose getting rid of it in both the twig and react versions of the library. This would be a breaking change for twig so perhaps it can be deprecated before getting removed. But I think encouraging consumers to use specific versions of the card components is the way to go.

justintemps:
Hey @bashlk, we don't have to have it in the React lib, but it's already being used in the Twig one. We could perhaps add a warning that it will be deprecated in future versions, but for now we have to support it.

Shashika6:
Hey @justintemps Cards are done 🎉  all cards are now individual components and have their own stories and props. 

Had a chat with @bashlk as well, The default card component is still being used by the cardgroup so if we remove it completely it will break. I think we should create another issue to fix the cargroup to use these individual components instead of the common card component.

justintemps:
Thanks @Shashika6, I still had to do a lot of cleaning up on this. See some comments below.

1. Only one card takes the `cornercut` prop as you'll see in the Twig Storybook [Twig storybook](https://twig.ui.ilo.org/?path=/docs/components-cards-card--default). You were passing it to two or three other cards that couldn't use it (Factlist Card and Data Card). In fact, toggling the property in Storybook didn't do anything so that might have rung some bells. 
2. In terms of typing, `cornercut` has only two options, corner cut or no corner cut, so it should be boolean, not a union of strings.
3. Not all of the components take the same combination of sizes. Some of them have more options, others have less. I realize this is confusing, you just needed to go through Twig components and see which options were being passed to each component. Here again, toggling between these values in the React Storybook had no effect, so that might have suggested something was off.
4. The `size` property was set to optional for all of the cards, but what happens if the developer doesn't pass one in? As a rule, when a prop is optional, you need to set a default.
5. There were no comments on the props for the individual cards so the props were essentially undocumented.

Could you have a last look at this before I merge?

Shashika6:
Hey @justintemps looks good, I was following the cards with the designs not the twig. In the designs a few cards have cornercut in them.

justintemps:
Thanks @Shashika6, the question on the cornercut isn't whether they have one or not, it's whether they have the option to choose.

Shashika6:
Right @justintemps , Sorry for the confusion

Issue #765, opened 2024-01-10
Modal: Elements with role="dialog" or role="alertdialog" do not have accessible names


justintemps:
I think some element in the Modal component just needs to be given a `name` attribute.

![Screenshot 2023-12-11 at 14 53 59](https://github.com/international-labour-organization/designsystem/assets/12401179/7263902d-e074-4629-9699-e12a30b8f98a)

⚠️ Evidently a name attribute won't do it, they need a aria-label attribute

PR #770, opened 2024-01-10
feat(twig): add aria-label attribute to modal

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/765

### Notes :- 

* Added aria label attribute to modal in twig.
* **Added modalLabel as an optional param as it would break the existing implementation**

### DIFF
--- a/.changeset/quiet-ears-applaud.md
+++ b/.changeset/quiet-ears-applaud.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": patch
+---
+
+Add optional string modalLabel parameter to be used as aria-label attribute in modal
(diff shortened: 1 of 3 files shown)


justintemps:
Hey @Shashika6 really sorry! I looked into this and the error is a little misleading, evidently what's actually needed here is an `aria-label` attribute. So may be the argument should be `modalLabel` or something? 

See more info here: https://dequeuniversity.com/rules/axe/4.7/aria-dialog-name

Shashika6:
Will do @justintemps I'll update the changes and add a comment

justintemps:
Thanks @Shashika6 could we be just a little more descriptive in Storybook about what the argument is for? It should be clear that's it's for accessibility.

Shashika6:
Updated the description @justintemps

Issue #772, opened 2024-01-11
Multilink card: image alignment does not work


justintemps:
From a website developer (@Shashika6 can you make sure this is accurate before implementing? I think it is, but just double check).

The issue lays with this line https://github.com/international-labour-organization/designsystem/blob/develop/packages/twig/src/patterns/components/cardgroup/cardgroup.twig#L27 

It should be passed as

```
alignment: alignment,
```

As the alignment is not a property of the card, but of the group. That way, it will pass over the value to the [card template](https://github.com/international-labour-organization/designsystem/blob/develop/packages/twig/src/patterns/components/card_multilink/card_multilink.twig#L5) to apply the corresponding class and styles will be properly applied.

Shashika6:
Sure @justintemps will test this out and do the change if its valid.

PR #774, opened 2024-01-11
fix(twig) - Alignment issue in cardgroup

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/772

### Notes :- 

* Verified and refactored alignment to use param from cardgroup.
* Added alignment as a storybook setting for cardgroup.


Shashika6:
@justintemps I added alignment as a setting as well, Let me know if thats okay or it should be removed

Issue #764, opened 2024-01-10
Lists do not contain only <li> elements and script supporting elements


justintemps:
It looks like this is happening in the List component, I think because the `<ul>` or `<ol>` has the heading nested inside of it and shouldn't. I think this comes from a lighthouse scan so should be straightforward to check.

![Screenshot 2023-12-11 at 15 49 54](https://github.com/international-labour-organization/designsystem/assets/12401179/8c573a8c-fe10-4b53-9baf-953b8a44f19e)

Shashika6:
Had a look at this @justintemps,

Tried two approaches for this :- 

* **Converted existing <h5> to <li>**

* This adds the list styling to the heading as well. 

* **Moved the <h5> out of <ul> and <ol>**

* This breaks the layout for horizontal.

Looks like we will have to do alot of changes to the style and layout as well @justintemps What do you think we should do ?

justintemps:
@Shashika6 can we not just wrap the whole thing in a <div> that you can nest the heading under?

PR #806, opened 2024-02-06
fix(react/twig) - Refactor list component to contain only li elements

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/764

### Notes :- 

* Refactor the list component to make sure that only li elements are children of parent lists.

### Changes :- 

* Refactor and ensure both twig and react follow the same structure


Shashika6:
@justintemps Moving out the title from the components breaks the styling had to refactor the styling a bit to make it work. Cc :- @bashlk

Issue #780, opened 2024-01-15
Tag: 'reset' status should not make the tag taller


inesdgomes:
'reset' tags look taller than all other tags' statuses. They should be the same height.

<img width="325" alt="Screenshot 2024-01-15 at 09 50 36" src="https://github.com/international-labour-organization/designsystem/assets/22772102/6c074d00-61fa-4801-aa5f-57663634c9c6">

See example here: https://iloorgstg.prod.acquia-sites.com/search?search_api_fulltext=france&ilo_date%5Bmin%5D=&ilo_date%5Bmax%5D=&sort_by=ilo_date&f%5B0%5D=language%3Aen&f%5B0%5D=language%3Aen 

I think the solution is to ensure the tag has the same height in all statuses and that the border on the 'reset' status shows inside rather than outside. This is the CSS code that Figma suggests:

```
border-radius: var(--2, 2px);
border: var(--0, 1.5px) solid var(--Greys-Grey-UI, #B8C4CC);
```

More info about the component: https://www.figma.com/file/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?type=design&node-id=5868-8576&mode=design&t=sZ2GlWgFlNp4yZqO-0

justintemps:
The problem here is in the Tag component, it should be the same height when `tagType` is `button` or `reset`.

By the way, these are terrible names, once this is done we should create a v2 ticket to change that.

PR #810, opened 2024-02-08
fix(tag) - Refactor reset tag padding to consider border

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/780

### Notes :- 

* Currently the reset button looks to be off when compared to the other buttons because it has a border.

### Fix :- 

* Consider the border width when setting the reset button inner padding so that it maintains consistency with other buttons.

### DIFF
--- a/.changeset/eighty-spoons-attack.md
+++ b/.changeset/eighty-spoons-attack.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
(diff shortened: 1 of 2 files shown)


justintemps:
Hey @Shashika6 this looks good to me, is it ready to be merged?

Shashika6:
Yes @justintemps its ready

Issue #809, opened 2024-02-08
Pagination component has irregular spacing forward with respect to gap


justintemps:
The pagination component has a gap between the forward and end buttons that's not there between the back and start buttons. 
See example here: https://twig.ui.ilo.org/?path=/docs/components-navigation-pagination--pagination

![Sem título-20240125-165807](https://github.com/international-labour-organization/designsystem/assets/12401179/abea5c7e-1f68-4d24-8d57-172b2a3f8794)

The spacing should be consistent.

PR #812, opened 2024-02-09
fix(style) : Remove gap between forward and end button

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/809

### Notes :- 

* There is a gap between the forward and end button. The reason this was added is to center the pagination count.

### Fix :- 

* Remove gap between forward and end button and center the pagination count.

### Screenshot :- 

Screenshot 2024-02-10 at 00 15 18

### DIFF
--- a/.changeset/sweet-feet-grow.md
+++ b/.changeset/sweet-feet-grow.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Remove gap between forward and end button in pagination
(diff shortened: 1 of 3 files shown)


justintemps:
@Shashika6 while we're here, I think we need to update the docs. "The pagination component displays a dismisable alert isn't correct." Could we change that to something that makes sense?

Shashika6:
Updated the description as well @justintemps

Issue #782, opened 2024-01-18
Breadcrumb gets hidden in small devices


inesdgomes:
@justintemps The breadcrumb gets hidden when the device is not large is this the expected behaviour ?

_Originally posted by @Shashika6 in https://github.com/international-labour-organization/designsystem/issues/571#issuecomment-1868749031_

inesdgomes:
@justintemps Sorry, saw that @Shashika6 mentioned this in another ticket and indeed it shouldn't be working like this. The breadcrumb should still be visible on mobile.

GGKapanadze:
How can I reproduce the hidden breadcrumb? because it works properly in the examples below
@inesdgomes @justintemps 


> Chrome Devtools, iPhone SE
![Image](https://github.com/international-labour-organization/designsystem/assets/36326203/f5bdb787-4b05-4cf5-b7b8-b5fa35a4ea30)

> Physical device, iPhone 15 pro

https://github.com/international-labour-organization/designsystem/assets/36326203/05d1fbad-59c2-41ed-8e78-927dd62a6314

justintemps:
@GGKapanadze Check the hero component, that's where this is happening.

Shashika6:
Picking this up as discussed @GGKapanadze

PR #820, opened 2024-02-14
fix(style) : Breadcrumb hidden in small devices

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/782

### Notes :- 

* Currently the breadcrumb is only displayed to larger devices in the hero component. 

### Fix :- 

* Remove the property which sets the breadcrumb to only be included in large devices

### DIFF
--- a/.changeset/slow-baboons-occur.md
+++ b/.changeset/slow-baboons-occur.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Fix breadcrumb getting hidden in small devices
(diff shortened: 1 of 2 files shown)


Shashika6:
@inesdgomes The fix to remove the breadcrumb only shown on large devices is done it looks fine in large mobiles/tabs but in very small devices it looks a bit off. There isn't a design for this but maybe we should reduce the padding in very small devices so it fits, What do you think ? Cc :- @justintemps 

Screenshot 2024-02-15 at 00 14 33

Issue #469, opened 2023-08-21
Fix spacing issues in various components


inesdgomes:
- [x] **Multilink card**: revise all spacings in the multilink card, including top, left, right and bottom padding; before/after heading; before/after image; before/after description text
- [x] **Stats card**: revise space between last card and CTA on desktop and mobile

<img width="847" alt="Screenshot 2023-08-21 at 12 17 31" src="https://github.com/international-labour-organization/designsystem/assets/22772102/276b1116-aac7-43fa-b463-0f1134f11257">

- [x] **Profiles**: align 'Learn more' button to name and title

![Image](https://github.com/international-labour-organization/designsystem/assets/22772102/da5883bf-78e5-48f6-b0ea-8526b7de1d1c)

- [x] **Card with short description**: fix space between eyebrow, title and description; fix space between image and grey line.

<img width="1404" alt="Screenshot 2023-10-03 at 14 34 50" src="https://github.com/international-labour-organization/designsystem/assets/22772102/85b55607-21ef-49ae-9aa4-2ef4f717e70c">

<img width="1372" alt="Screenshot 2023-10-03 at 14 56 11" src="https://github.com/international-labour-organization/designsystem/assets/22772102/31bd5d99-2003-489c-ba8d-e8d41ce2cd2c">

justintemps:
@inesdgomes could you please update this ticket?

inesdgomes:
@justintemps I updated this ticket. I suggest you tackle the tickets below first as they may partially resolve #469. Once the ones below are addressed, we can come back to this one. 

[Cards' category](https://github.com/orgs/international-labour-organization/projects/2/views/3?pane=issue&itemId=44127331)
[Content category](https://github.com/orgs/international-labour-organization/projects/2/views/3?pane=issue&itemId=44127853)
[Feedback category](https://github.com/orgs/international-labour-organization/projects/2/views/3?pane=issue&itemId=44128254)
[Forms category](https://github.com/orgs/international-labour-organization/projects/2/views/3?pane=issue&itemId=44128469)
[Media category](https://github.com/orgs/international-labour-organization/projects/2/views/3?pane=issue&itemId=44128574)
[Navigation category](https://github.com/orgs/international-labour-organization/projects/2/views/3?pane=issue&itemId=44128705)
[Transitions and user interface categories](https://github.com/orgs/international-labour-organization/projects/2/views/3?pane=issue&itemId=44128893)

Shashika6:
[...]

PR #822, opened 2024-02-17
fix(style) : Spacing issues in cards

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/469

### Notes :- 

Verified all of the cards

- [x] **Multilink card**: revise all spacings in the multilink card, including top, left, right and bottom padding; before/after heading; before/after image; before/after description text

- [x] **Stats card**: revise space between last card and CTA on desktop and mobile

- [x] **Profiles**: align 'Learn more' button to name and title

- [x] **Card with short description**: fix space between eyebrow, title and description; fix space between image and grey line.

### DIFF
--- a/.changeset/late-worms-live.md
+++ b/.changeset/late-worms-live.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Refactor spacing in profile and cardgroup to use new mixin
--- a/packages/styles/scss/components/_cardgroup.scss
+++ b/packages/styles/scss/components/_cardgroup.scss
@@ -60,7 +60,7 @@
   &--button-wrap {
     display: flex;
     justify-content: center;
-    margin-top: px-to-rem(64px);
+    margin-top: spacing(16);
     width: 100%;
   }
 }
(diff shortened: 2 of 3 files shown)


Shashika6:
Screenshot 2023-10-03 at 14 56 11

justintemps:
hey @Shashika6 correct me if I'm wrong, but here it looks like we're just updating the same spacing values using the new `spacing` mixin, right?

Shashika6:
Yes @justintemps for now its like that waiting on the comment above to proceed with it

Shashika6:
@justintemps I think we can merge this, verified that the spacing is in sync with the design as well. The changes are using the new spacing function.

justintemps:
@Shashika6 just to confirm, the only thing this does is uses the new spacing mixin, it doesn't actually change any of the spacing right?

Shashika6:
Yes @justintemps thats correct

justintemps:
Ok @Shashika6 then I think we should change the changeset message, because it doesn't actually fix any spacing issues, it just uses a mixin to apply the same sizes that were there before.

Shashika6:
Updated the changeset message @justintemps

PR #829, opened 2024-02-22
Twig Icon Component

GGKapanadze:
# Description

Introducing a new icon component for twig 

https://github.com/international-labour-organization/designsystem/assets/36326203/293ae5eb-2005-4615-a234-e755f880274d

### API
The icon component accepts 3 props

- **name** - name of the icon, any icon from `@ilo-org/icons` package, case sensitive. [Full list](https://github.com/international-labour-organization/designsystem/tree/develop/packages/icons/src/svg)
- **size**  - the fixed size of an icon, can be: `16`, `20`, `24`, or `32`. Optional
- **color** - any arbitrary CSS color value: `color`, `hex`, `rgb`, `rgba`, `hsl`. Optional

so twig template will look like this 
```twig
{% include "@components/icon/icon.twig" with {
  "name": "Facebook",
  "size": "32",
  "color": "hsl(50 80% 40%)"
} %}
```

### Implementation

As discussed in #692, the Icon component is done via the `js` implementation, icon component has a dependency on `@ilo-org/icons`. The first render includes an empty `SVG` element `24x24`, and js replaces it with a new SVG based on the user's input. There are 2 edge cases,
1.  when given an incorrect icon size the system throws a warning and defaults to `24px`
image

2. Incorrect Icon name is given
image

The final `SVG` retains the `ilo--icon` class name and `js-processed` data attribute 
image

### Bundle 

The js logic has no overhead or size problem, bundle is increased by the size of `@ilo-org/icons` 
image
 

### Drupal Demo

Here is a demo from the Drupal website using a new icon component

https://github.com/international-labour-organization/designsystem/assets/36326203/436c16da-2355-4f82-b844-b10c13b2992e

```twig
{% include "@components/icon/icon.twig" with {
  name: 'Facebook',
  size: 32,
  color: '#316FF6'
 } %}

{% include "@components/icon/icon.twig" with {
  name: 'Arrow',
  size: 24,
  color: 'red'
 } %}

{% include "@components/icon/icon.twig" with {
  name: 'Home',
  size: 32,
  color: '#316FF6'
 } %}
```

Issue #692, opened 2023-11-23
Twig: refactor Icon component 


justintemps:
## Description

Right now the icon component in Twig looks like this:

```
<img class="{{prefix}}--icon" src="{{'/images/'~icon~'.svg'}}" alt="{{icon}}" role="presentation">
```

This is sort of a naive implementation that assumes there's some top-level `/images` directory that has all of the icons in the Design System. That's true in the Twig package itself, and so it works in Storybook, but it's not the case in most Drupal Websites.

We need to refactor this component so that it's either using CSS to render the icons as background-images from `@ilo-org/icons/build/icons.scss` or we need to find the standard Twig way of telling the server that the path to the icons is actually in `/@ilo-org/icons/svg`

## Example 

See the mobile version of this page: https://iloorgstg.prod.acquia-sites.com/node/56

![image](https://github.com/international-labour-organization/designsystem/assets/12401179/bb629301-e27d-4225-bdb5-e1a2a426512a)

justintemps:
Let's explore the option of having a simple script that a Drupal developer can run in order to place all of the image files from the token package into a place where the Token component will be expecting to find them. We may also need to add a path arg to the Token component so that the Drupal development can keep icons somewhere other than `/images`.

GGKapanadze:
## Introduction

The new `icon` component should be done without any directory generation or "run an extra script" step for twig implementation. In the current context, we have 2 main ways 
- web fonts
- SVG sprite

tl;dr web fonts will 
- require complex build steps for fonts
- add additional effort for colorized icons
- make debugging more complex

IMO, SVG sprite is the good for our scenario. The main idea of Sprite is that we define one `svg` element and using `symbol`-s we can add multiple items inside of `defs`, after that we can reference any symbol with `use` keyword and `id` on each `symbol`.  

https://github.com/international-labour-organization/designsystem/assets/36326203/6acb4b61-b164-43c7-a91f-292345d9cc03


`svg` sprite is almost perfect in our case but it will require us to have some idea how are we doing the string concatenation for the sprite, aka how to unionize all icons into one big `svg`.  While I was playing with the concatenation problem with `@ilo-org/icon` build tools, I noticed that we have an `umd` export that contains all icons with all sizing bundled in one file as plain Javascript objects. It looks exactly like the sprite I mentioned above but with objects, so instead of `use` keyword we will utilize `JS` to create SVG elements from the exported format

[...]

PR #834, opened 2024-02-29
Twig: Fix Table of Contents close button not showing on mobile

### LINKED ISSUE #604: Sticky navigation close icon not rendered on mobile
The ‘'close’' icon of sticky navigation on mobile devices is not rendered or broken. Happens on android and iOS.

Example here: https://iloorgstg.prod.acquia-sites.com/node/611

firefox_GehTTFswJ0(1)
close(1)

justintemps:
This fixes #604 

This was caused by problems stemming from how icons were handled in the Twig package (described in #692). The root issue was eventually resolved in #829, however the Table of Contents still wasn't rendering the close icon in its mobile view correctly. 

There were two reasons for this:

1. The underlying Button component wasn't passing the right properties for the new Icon component
2. The ToC wasn't passing the right properties to the Button component

In fact, the Table of Content component was actually letting the user pass any properties to the Open & Close buttons that they wanted. This was especially problematic given that the Buttons will only work if they have a specific classname. The user shouldn't have to figure out what that classname is or be able to set it arbitrarily.

To correct these issues, the following was done:

1. Refactor the Button so that it passes the correct props to the Icon
2. Refactor the Table of Contents so that it only passes a subset of properties to the underlying Buttons it uses, ensuring they don't break.

### Before

image

### After

image

## 🎉 Bonus

This also includes two convenience scripts for preparing the Twig and React packages for development.

`pnpm react:dev:prep` and `pnpm twig:dev:prep` will each perform a clean install of their dependencies and then built the internal packages in the monorepo they rely on so you can simply run `pnpm twig:dev:docs`, for example, to start developing.

### DIFF
--- a/.changeset/many-mails-punch.md
+++ b/.changeset/many-mails-punch.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": patch
+---
+
+Enable Button to use newly refactored Icon component
--- a/.changeset/neat-trains-study.md
+++ b/.changeset/neat-trains-study.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": patch
+---
+
+Correct the way the Table of Contents passes properties to the Open and Close buttons.
--- a/contributing.md
+++ b/contributing.md
@@ -135,6 +135,12 @@ Build React Storybook and dependent packages
 pnpm react:build:docs
 ```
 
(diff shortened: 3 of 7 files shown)


justintemps:
Hey @GGKapanadze, thanks for the rimraf suggestion, that's done in #837

PR #844, opened 2024-03-04
Fix(themes): add fallbacks for Japanese fonts

### LINKED ISSUE #818: Japanese font: need to serve NotoSansJP
It seems we're serving NotoSans SC or NotoSans CJK fonts in the Japanese interface. This means the interface is showing Chinese characters rather than Japanese characters. We need NotoSansJP to show on Japanese interfaces.

justintemps:
Add Hiragino Sans and Yu Gothic as fallbacks for Noto Sans CJK JP (Japanese)

Fixes #818

### DIFF
--- a/.changeset/nervous-lies-return.md
+++ b/.changeset/nervous-lies-return.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/themes": patch
+---
+
+Add Hiragino Sans and Yu Gothic as fallbacks for Noto Sans CJK JP (Japanese)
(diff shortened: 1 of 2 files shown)


justintemps:
Hey @inesdgomes, I've added the fallbacks in Japanese for Noto Sans, could you help me test it? Here's what I'm seeing on my machine.

Screenshot 2024-03-04 at 15 11 33


Could you get a Japanese speaker to go here: https://deploy-preview-844--ilo-ui-twig-develop.netlify.app/?path=/story/components-content-rich-text--rich-text

And then paste a bunch of text in the content window at the bottom?

Issue #869, opened 2024-03-12
React : CardGroup is still using the common component 


Shashika6:
The cardgroup in react is still using the common card component.

The cards are now using individual components so we have to refactor the card group to use those components instead.

Shashika6:
@justintemps The card group is still using the common card component. Will do the changes to use the individual cards

PR #871, opened 2024-03-14
fix(react) - Refactor cardgroup to use individual component

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/869

### Notes :- 

The cardgroup component design is broken and it needs to use the individual card components instead of the common card component

### Fix :- 

* Sync classnames with twig
* Use individual card components

### Screenshots :- 

Screenshot 2024-03-14 at 13 42 19


justintemps:
Nice work @Shashika6

packages/react/src/components/Cards/CardGroup.props.ts
```diff
@@ -1,12 +1,31 @@
...
+  | StatCardProps
+  | TextCardProps;
 
 export interface CardGroupProps {
   /**
    * An array of card objects
    */
-  cards?: Required<CardProps>[];
-
+  cards: CardProps[];
```
bashlk, line 28: This type declaration will also allow different card types to be used at once. The CardGroup is only supposed to be used with a single type of card at once right?

packages/react/src/components/Cards/CardGroup.tsx
```diff
@@ -2,40 +2,53 @@ import { FC } from "react";
...
+  DataCard,
+  DetailCard,
+  FactlistCard,
+  MultilinkCard,
+  PromoCard,
+  StatCard,
+  TextCard,
+} from ".";
+
+const cardMapper: Record<string, FC<any>> = {
```
bashlk, line 27: Why FC<any> and not simply FC?

packages/react/src/stories/CardGroup/CardGroup.mdx
```diff
@@ -9,7 +9,7 @@ import { CardGroup } from "../../components";
 Card Group displays a group of Cards.
 
 <Canvas>
-  <Story of={CardGroupStories.GraphicCardGroup} />
+  <Story of={CardGroupStories.FeatureCardGroup} />
```
bashlk, line 12: Why change the default story?

bashlk (changes requested):

packages/react/src/components/Cards/CardGroup.props.ts
Shashika6: Yes @bashlk, I added a union type here to make sure that the mapper card component will have one of the above types any suggestions ?

packages/react/src/stories/CardGroup/CardGroup.mdx
Shashika6: Changed it to be in the same order as twig, in the twig its defaulting to feature card.

packages/react/src/components/Cards/CardGroup.tsx
Shashika6: Hmm didn't get you here @bashlk

packages/react/src/components/Cards/Card.props.ts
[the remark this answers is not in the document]
Shashika6: Will do @bashlk

packages/react/src/components/Cards/CardGroup.tsx
bashlk: Uff Github ate my angle brackets 😅 - should make more sense now.

packages/react/src/components/Cards/CardGroup.props.ts
bashlk: I mean something like this

```
export interface CommonCardGroupProps {
  cardCount: Required<string>;
  cta?: LinkProps;
}

export interface StatCardGroup extends CommonCardGroupProps {
  type: "stat";
  cards: StatCardProps[];
}

export interface MultilinkCardGroup extends CommonCardGroupProps {
  type: "multilink";
  cards: MultilinkCardProps[];
}

...

export type CardGroupProps =
  | StatCardGroup
  | MultilinkCardGroup
  | TextCardGroup
  | PromoCardGroup
  | FeatureCardGroup
  | DetailCardGroup
  | FactlistCardGroup
  | DataCardGroup;
```

With this, someone won't be able to provide cards of a different type than the one specified as the type to the CardGroup.

packages/react/src/components/Cards/CardGroup.tsx
bashlk: Also on a second look cardMapper doesn't need to declare components / functions - you can have it return the components directly

e.g.
```
const cardMapper = {
     stat: StatCard,
     multilink: MultiLinkCard,
     ...
}
```

packages/react/src/components/Cards/CardGroup.props.ts
bashlk: Also note that card count right now is not camel case. It probably should be. 

Also I don't think it is required at all. We can count the number of items passed into the cards prop to get the number of cards.

Shashika6: The cardCount here refers to the amount of cards per row @bashlk. I think it makes sense to refactor the name of that variable right ?

Shashika6:
@justintemps This is ready now, @bashlk made the changes you requested as well

justintemps:
Hey @Shashika6, I think you still need to add the doc block quotes so they'll appear in the `prop` descriptions. You do this literally just by adding comments to the props in the `props.ts` files. See `Card.props.ts` for an example.

Shashika6:
Hey @justintemps added the prop descriptions as well.

Issue #357, opened 2023-05-31
React/Docs: Stories should be appropriately laid out


justintemps:
Rather then wrapping them in elements, this should use the `meta.parameters.layout` property in the Stories.

PR #873, opened 2024-03-16
fix(react) - add paramater layout to storybook

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/357

### Notes :- 

* Currently there is no padding set in stories and padding is set to 0

### Fix :- 

* Add parameter layout padded
* Remove padding reset to 0

### Before

Screenshot 2024-03-16 at 18 12 08

### After

Screenshot 2024-03-16 at 18 12 34

### DIFF
--- a/.changeset/seven-ducks-rescue.md
+++ b/.changeset/seven-ducks-rescue.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/react": patch
+---
+
+Add paramater layout to storybook and remove padding reset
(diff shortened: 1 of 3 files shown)


justintemps:
Thanks @Shashika6 can you make sure that we then remove any wrapper components that do nothing but add padding?

Shashika6:
Sure @justintemps looks like the components do not have any unwanted wrapper components but i did see some unwanted styles in the preview-head.html shall we clear those out too ?

justintemps:
@Shashika6 yea that sounds like a good idea

PR #875, opened 2024-03-19
🌿 TWIG Home Screen 

GGKapanadze:
# Description 

This PR introduces a new home screen for the twig design system and also adds an indicator badge to the sidebar

image

image

## Design
There are two minor adjustments to the design

1. Logo: The original twig logo is too high to display on the left side, so I moved it to the right. The text is adjusted on the left side and the logo is trying 😀 to wrap them from the right
2. Introduced a more dense style for the version indicator

image

## Animation 

I have experimented with a variety of triangle animations but finished with a small triangle spinning upon hover, it was the most settled one

https://github.com/international-labour-organization/designsystem/assets/36326203/58e4fa25-f747-48f1-96aa-5f0a51b67dca

## Responsive 
Triangle responsiveness is also implemented, on mobile triangles aren't moving 

image

## React

After discussed about twig implementation I will do follow up for react


Shashika6:
Just excited to see this good work @GGKapanadze 🎉

justintemps:
This is nice @GGKapanadze ! Just a few things.

First, apologies, here is the Figma file for the original concept: https://www.figma.com/file/qaufMvHrykMaK69TAOEJ0D/Untitled?type=design&node-id=0%3A1&mode=design&t=XyrsBwUoLIIpNlgs-1

- I think we should use the icon as with the Figma designs, it's a bit more subtle.
- OK to use smaller text for the version number but I think we avoid italics as a rule
- The animations aren't quite there, I was thinking all three of the triangles would move horizontally in different directions (may be one left, the other right, the other left) but really slowly, almost imperceptibly. Should be very subtle. What do you think?

GGKapanadze:
Updated UI + animation

https://github.com/international-labour-organization/designsystem/assets/36326203/a9e95d63-79d2-495a-9916-7a4da4541d94

Issue #874, opened 2024-03-18
Twig: clean up npm package


justintemps:
Here is a screenshot of what's include in the `@ilo-org/twig npm pacakge`:

![image](https://github.com/international-labour-organization/designsystem/assets/12401179/9e12e335-ba8f-4912-b5f7-61875bfb3934)

## Things that should be there but aren't

`/dist` for some reason isn't getting included. I think this is because it's in `.gitignore`, simply adding a `.npmignore` file should fix this.

## Things that are there but shouldn't be

To double check, but everything else. Nothing else is needed except for the dist folder.

## ⚠️ This will probably be a breaking change for ilo.org

Given that we haven't been exporting the `/dist` folder 🤦 the developers over there must be getting assets from the `/src` folder. So we'll have to correct that before we merge this.

As an intermediate step, we should leave the `/src` folder for now so that we don't break anything, but add a note to deprecate getting assets from there later on.

bashlk:
This was fixed in #880. I think we should keep this either keep this ticket open or open another one so that we remember to stop publishing the src folder once the ilo.org devs have moved to using the dist folder.

Issue #762, opened 2024-01-10
Export Styles package from Twig and React packages


justintemps:
Right now, we're forcing users to import at least two packages which are both versioned separately if they want to use either component library.

Example, if I'm using the Twig library right now I have to have.

- @ilo-org/twig@0.14.2
- @ilo-org/styles@0.13.1

It's normal for them both to be versioned separately, because not all changes to Twig will affect styles and vice-versa, but we're still making users figure out which versions they need of the two libraries. This is going to be very confusing if you're on an older version and aren't ready to update to the newest one yet.

To solve this, we should export the Styles package directly from the Twig and React packages. They already have it as a dependency, we could just add a build stage where it gets added as an exported package.

That way we could go from this:

```tsx
import { Button } from "@ilo-org/react";
import "@ilo-org/styles/css/button";
```

To something like this:

```tsx
import { Button } from "@ilo-org/react";
import "@ilo-org/react/styles/button";
```
 
You go from having to manage two dependencies as a user to one.

bashlk:
@justintemps Both of the example snippets provided are the same 😄

justintemps:
> @justintemps Both of the example snippets provided are the same 😄

Nope, there's a subtle difference. Look carefully.

bashlk:
Ah my bad, I see it now - it would be better if we don't make the user import the styles in the first place, so that they import only the Button component in this example.

justintemps:
It would but that would make some assumptions about the way the user is
bundling things no?

bashlk:
I don't think so - the styles of the Button are a concern of the design system. 

At least on React, I think we should hide this from the user as much as possible so that they can just import the component and use it without having to manually import the styles as well. I have not used any component library that has made me manually import the styles for each component.

A simple way to achieve this is to add an installation step asking the user to import a stylesheet containing all the styles for the designstem to their index.js / index.ts file (something similar to packages/styles/scss/monorepo.scss). That way with one step, they have all the styles setup for the design system.

The downside of this is that the stylesheet will contain all of the styles, even the styles for components that the user is not using. But given the added convenience, I think the cost will be worth it. Perhaps we can break it down into groups, so that the user can import the stylesheet for a group of components. But even that might be too complicated than required.

[...]

PR #880, opened 2024-03-19
fix(core): reorganized package structure

### LINKED ISSUE #876: Twig: Reorganize how components are exported in the package
This extends [#762](https://github.com/international-labour-organization/designsystem/issues/762)

Each Twig component should include:

- Twig file
- Pattern file (yml)
- Bundled behaviour file (.js)
- Stylesheet for that component

This needs to be done in a way that supports the legacy implementation (where assets are retrieved directy from `/src/`)

bashlk:
Fixes #876 and #874 

Previously the dist folder was not being published to npm and it didn't contain the twig / wingsuit components in a format that was ready to be consumed by Drupal.

This PR changes the structure of the dist folder such that the everything needed by Drupal to render a component is placed within its own folder. (i.e. the twig template, the wingsuit definition in yaml, the bundled javascript and the compiled css for a component are placed in a folder named after the component) So this also addresses #762 for the twig package since the compiled styles are now inside the twig package and the end user does not need to import and use the styles package seperately.

This PR also introduces a .npmignore file to make npm publish the dist folder and to avoid it from publishing the storybook and other dev tool configurations.


packages/twig/package.json
```diff
@@ -41,15 +40,19 @@
     "@ilo-org/styles": "workspace:*",
     "@ilo-org/themes": "workspace:*",
     "@ilo-org/utils": "workspace:*",
+    "@popperjs/core": "^2.11.8",
     "video.js": "^7.21.2",
-    "videojs-youtube": "^3.0.1",
-    "@popperjs/core": "^2.11.8"
```
justintemps, line 46: @bashlk we should probably leave this in alphabetical order, no?

justintemps:
Hey @bashlk, I just ran `npm pack` in the Twig directory and this is the structure I'm seeing:

```
├── CHANGELOG.md
├── LICENSE
├── README.md
├── directory-structure.txt
├── package.json
└── src
    ├── namespaces.js
    ├── patterns
    │   └── components
    ├── tokens
    │   ├── exports
    │   └── index.scss
    └── vendorjs
        └── lib.vendor.js
```

packages/twig/package.json
bashlk: I didn't rearrange the order of package, pnpm did 😄

bashlk:
@justintemps Apparently the `npm pack` command doesn't work with pnpm packages. We should run `pnpm pack` instead. Changesets apparently uses `pnpm pack` automatically when it detects it - https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish

justintemps:
@bashlk that's correct, but even with `pnpm pack` I'm not seeing the `dist` folder.

justintemps:
@bashlk, I add `!dist` to the `.npmignore` file in the Twig package and that seemed to work. I think `pnpm pack` might check both the `.npmignore` and `.gitignore` files and if a file is specified in either one, then it won't added it to the tarball.

packages/twig/package.json
justintemps: That's really weird. I'm not going crazy, `@` packages should come first right?

ademarco:
@bashlk not sure it's related to the current PR but it didn't seem to happen on develop. When I build the design system I get the css of some components in their own directory, and not alongside the component:

image

Maybe this is due to the fact that the name of the component, in the YML file, (`datacard`) does not match the name of the component file and directory (`card_data`)? There are several such cases.

justintemps:
hey @bashlk I just pushed up a few commits that 

1. Remove all of the `.npmrc` files throughout the project
2. Add a `files` property to the root package.json

This has the desired effect of packing the modules correctly with all of the `dist` assets that we need. Have a look and give it a try to make sure you agree with what I've included.

⚠️ I did not include the Twig `/src` folder. I checked the ILO.org Drupal project and the `/src` path is only getting used in two places, so I think we can easily make a PR to that project when we're ready with a new version of the Design System and the correct path for the Twig assets.

bashlk:
@justintemps The current wildcard glob patterns for the LICENSE, README.md and more are also matching files inside the node_modules folders of the individual packages. We shouldn't include node_modules in the published assets. Also the package.json files for all of the packages are missing.

I think we should also remove the useless files properties inside each packages' package.json file. I already had this change locally so I pushed it up just now.

justintemps:
@bashlk good catch. I'll finish working on the packages on this branch since we're already here. I'll then leave you to make any changes to the Twig package.

bashlk:
@ademarco Yes, the styles are being placed in the wrong folder because the filenames don't match (See copystyles.js) - previously all of the CSS files were being placed in a single folder, I changed it so that they are being placed alongside the other component files. Do you think the previous approach was better?

ademarco:
> @ademarco Yes, the styles are being placed in the wrong folder because the filenames don't match (See copystyles.js) - previously all of the CSS files were being placed in a single folder, I changed it so that they are being placed alongside the other component files. Do you think the previous approach was better?

We can actually roll back to the previous approach, as I realized that the current wingsuit definition do not properly declare dependencies between patterns and, in failing to do so, we are forced to load the CSS of all components, or styles will break.

If possible, I would keep the new approach of having the JS compiled in .behaviour.js alongside the twig files tho, as those can be loaded conditionally.

justintemps:
@bashlk ok, new glob pattern in the root package.json adds package.json and removes node_modules. All the packages are now significantly smaller. 🎉

Issue #773, opened 2024-01-11
Apply margin to <iframe> elements in RichText styles


justintemps:
See the example below, we're applying the appropriate amount of space to images in the RichText component, but we need to add a refule for `<iframe>` elements too so the same will apply to embeds.

<img width="1649" alt="Screenshot 2024-01-10 at 16 26 00" src="https://github.com/international-labour-organization/designsystem/assets/12401179/ea41eb88-71aa-492b-b7ab-818838dcb3ba">
<img width="1628" alt="Screenshot 2024-01-10 at 16 25 51" src="https://github.com/international-labour-organization/designsystem/assets/12401179/9aa1dbf4-b426-4bb0-8dc9-9e4765d62a4e">

justintemps:
Hey @Shashika6, re-opening this because it seems like the issue persists in situations like this one: 

https://iloorg.prod.acquia-sites.com/resource/news/template-global-unemployment-rate-set-increase-2024-while-growing-social

cc @inesdgomes

justintemps:
@inesdgomes, can you provide a different link, we can't seem to be able to access this one:

https://iloorg.prod.acquia-sites.com/resource/news/template-global-unemployment-rate-set-increase-2024-while-growing-social

justintemps:
Hey @Shashika6, the link in the example is available now, you just need to use the basic auth password that you have for now: https://iloorg.prod.acquia-sites.com/resource/news/template-global-unemployment-rate-set-increase-2024-while-growing-social

justintemps:
Hey @Shashika6, could you tackle this one next? Make sure to look at it with bb8 as we've been back and forth with this one a couple of times already.

Shashika6:
Hey @justintemps the link doesn't seem to work for me it keeps saying forbidden access :\

inesdgomes:
Hi @Shashika6 Can you check now?

Shashika6:
Works now @inesdgomes thanks :)

PR #886, opened 2024-03-21
fix(style) - Add spacing to iframe in richtext

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/773

### Notes :- 

* Issue still exists because some iframes are added with `article` elements instead of `p`. 

### Fix :- 

* Move out spacing to iframe element so that it would work in any element.


justintemps:
Hey @Shashika6 I just tested that using [bb8](https://github.com/international-labour-organization/designsystem-site-builder) and it didn't work, it looks like you need to apply the `margin-bottom` to the `article` here, not the `iframe`.

Just adding `article` here should do it:

https://github.com/international-labour-organization/designsystem/blob/a69e6480657dcdf88655ade3b60a134116b1cea5/packages/styles/scss/components/_richtext.scss#L18

packages/styles/scss/components/_richtext.scss
```diff
@@ -333,6 +331,10 @@
       );
     }
 
+    iframe {
```
justintemps, line 334: I don't think this is doing anything, there's no `figure` in this example that I can see.

Shashika6: I got it confused here @justintemps since the initial issue was about adding spacing to iframe. Will make the change above and test it

PR #897, opened 2024-03-27
#895 Fix wingsuit YAML definitions

### LINKED ISSUE #895: Make Wingsuit YAML definitions compatible with UI Patterns
As part of our work on the [ILO Base Theme](https://github.com/international-labour-organization/ilo_base_theme) we are trying to have UI Patterns read the component definitions contained in `.wingsuit.yml` files. Unfortunately many of such definition do not validate, for several reasons, for ex. PHP YAML parser getting picky about typing, values referring to non existing plugins in Drupal, etc.

The scope of this ticket is to fix the definitions, so that they can be actually used in a Drupal site.

ademarco:
Fixes #895

### DIFF
--- a/packages/twig/src/patterns/components/button/button.wingsuit.yml
+++ b/packages/twig/src/patterns/components/button/button.wingsuit.yml
@@ -71,8 +71,8 @@ button:
       description: The disabled state of the button.
       required: false
       options:
-        true: "true"
-        false: false
+        "true": "true"
+        "false": false
       preview: "false"
     kind:
       type: select
@@ -98,8 +98,8 @@ button:
       label: Icon Only
       description: Is this an icon-only button?
       options:
-        true: "true"
-        false: false
+        "true": "true"
+        "false": false
       preview: false
       required: false
     opensmodal:
@@ -107,8 +107,8 @@ button:
       label: Opens Modal
       description: Does this button trigger open a modal?
       options:
-        true: "true"
-        false: false
+        "true": "true"
+        "false": false
       preview: false
       required: false
   visibility: storybook
--- a/packages/twig/src/patterns/components/button/button.wingsuit.yml
+++ b/packages/twig/src/patterns/components/button/button.wingsuit.yml
@@ -72,7 +72,7 @@ button:
       required: false
       options:
         "true": "true"
-        "false": false
+        "false": "false"
       preview: "false"
     kind:
       type: select
@@ -99,8 +99,8 @@ button:
       description: Is this an icon-only button?
       options:
         "true": "true"
-        "false": false
-      preview: false
+        "false": "false"
+      preview: "false"
       required: false
     opensmodal:
       type: select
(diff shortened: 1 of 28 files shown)


packages/twig/src/patterns/components/button/button.wingsuit.yml
```diff
@@ -71,8 +71,8 @@ button:
       description: The disabled state of the button.
       required: false
       options:
-        true: "true"
-        false: false
+        "true": "true"
```
bashlk, line 74: The true value in this case is a string rather than a boolean. I think all these should be booleans right?

bashlk: There are several more instances of this. and a minor thing which is that true / false is sometimes capitalized and other times now 😆

packages/twig/src/patterns/components/icon/icon.wingsuit.yml
```diff
@@ -5,7 +5,7 @@ icon:
   description: A component for displaying an svg icon.
   settings:
     name:
-      type: string
+      type: value
```
bashlk, line 8: What is the difference between type string and type value?

packages/twig/src/patterns/components/pagination/pagination.wingsuit.yml
```diff
@@ -9,7 +9,8 @@ pagination:
       label: Current page
       description: current page number that the user is on, zero-indexed
       required: true
-      preview: "0"
+      preview: 0
+      escape: false
```
bashlk, line 13: What does escape do?

packages/twig/src/patterns/components/button/button.wingsuit.yml
ademarco: yes, it's all a bit messy, I'll fix this by making them coherent, I'll make the labels as strings, with capital letters, so like this:

```
        "true": True
```

packages/twig/src/patterns/components/icon/icon.wingsuit.yml
ademarco: On the Drupal site, those are mapped to Drupal plugin IDs, and the plugin `string` does not exist, while `value` does.

packages/twig/src/patterns/components/pagination/pagination.wingsuit.yml
ademarco: That tells Drupal to not escape the preview value, otherwise the `0` is turned into a `Markup` object with value 0, which then breaks the preview since we then use that preview value in an arithmetic operation, like: `currentPage + 1`.

justintemps:
@ademarco so this is good, can we merge now?

ademarco:
@justintemps  there was one leftover on the options, I fixed it and forced pushed on the same commit, as it belonged there, we can merge it if @bashlk  is ok with it

bashlk:
@ademarco @justintemps Sure we can go ahead with the merge. I also made a PR that fixes the prefix substitution (#905) so we could also merge that and release all of this once.

Issue #865, opened 2024-03-11
Accessibility: address issues with headings' order in cards


inesdgomes:
@justintemps To be discussed:

- Feature cards: remove all heading tags; links should just be links.
- Graphic promo: change title from H3 to H2.
- Factoids: remove all heading tags, keep just standard paragraph text and one link per card.
- Multilink card: this is a complicated one because the headings should be kept as headings, but the level may change depending on whether the cards' group has a heading or not.

justintemps:
here's what we agree to do:

- Change the `h3` that's hardcoded as the title of the cards to a `p` by default
- Add a `headingLevel` prop that will let the developer use a different heading level than `p` if they need to.

The appearance of the title doesn't change, only the element that's rendered.

## Caveat

We understand that we may have to involve configuring the Promo Card (known as Highlighted Card on the website) so that it takes a H2 instead of the default P

justintemps:
Hey @inesdgomes and @Shashika6 on second thought, I think all of the cards should default to being a `p`. It doesn't make sense to have one default to being an `h2`.

PR #900, opened 2024-03-28
fix(react/twig) - refactor headings to improve accessiblity

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/865

### Notes :- 

* Currently h3 tags are used in cards which causes issues in accessibility in order of headings

### Solution :- 

* Refactor `h3` tags to be` p` tags by default.
* Add new prop called titleElement if a heading element needs to be added
* Refactored styling of promo card title to h2


justintemps:
Hey @bashlk in the React package, you need to fix the options provided for Storybook. It should be like a dropdown with a list of the supported elements. Right now it's set to be a generic object which is not right.

image

justintemps:
@Shashika6 the above comment goes for the Detail Card. The Factlist and Multilink Card don't seem to have the option in Storybook, which is a mistake.

Shashika6:
Hey @justintemps added a dropdown to select titleElement options in both twig and react.


Screenshot 2024-04-02 at 21 49 56

.changeset/small-brooms-destroy.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ilo-org/react": patch
+"@ilo-org/styles": patch
+"@ilo-org/twig": patch
+---
+
+Address accessibility issues with heading order in cards by changing `<h3>` tags to `<p>`, Additionaly add an optional prop that allows for configuration of the title element if needed.
```
justintemps, line 7: I would change this to:

Address accessibility issues with the heading order of cards by changing the default element used to render the title from an `<h3>` to `<p>`. Additionaly add an optional prop that allows for configuration of the title element if needed.

packages/react/src/types/index.ts
```diff
@@ -64,6 +64,7 @@ export type CardTypes =
   | "factlist"
   | "data";
 
+export type HeadingTypes = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```
justintemps, line 67: @Shashika6 This should include `"p"` if that's the default for cards.

packages/react/src/stories/Card/DetailCard.stories.tsx
```diff
@@ -17,6 +17,10 @@ const CardMeta: Meta<typeof DetailCard> = {
       options: ["wide", "narrow", "fluid"],
       control: { type: "radio" },
     },
+    titleElement: {
+      options: ["h1", "h2", "h3", "h4", "h5"],
```
justintemps, line 21: @Shashika6 for all cards, this should include "p" or else the user won't be able to return to the default.

packages/twig/src/patterns/components/card_detail/card_detail.wingsuit.yml
```diff
@@ -69,3 +69,14 @@ detailcard:
       options:
         "true": "true"
         "false": "false"
+    titleElement:
+      type: select
+      label: Title Element
+      description: HTML element used for the title
+      required: false
+      options:
+        h1: h1
```
justintemps, line 78: @Shashika6 here again, the options need to include the default which is "p"

justintemps (changes requested):
Hey @Shashika6 this looks mostly good, we just need to make sure that "p" is included among the options provided to the user.

Shashika6:
Made the changes you requested @justintemps :)

justintemps:
Hey @Shashika6, I added some more things here:

- Changed `titleElement` to `titleLevel` which I think is clearer
- Allowed the `titleLevel` to be set from the Card Group for all cards in the group

Which I was doing the later, I noticed that the Card Group in React still needs to be synced with the Card Group in Twig. There are some settings that are missing (alignment etc.). Could you open a ticket and take care of that?

Shashika6:
Will do @justintemps 🫡

PR #901, opened 2024-03-28
Styles: Refactor List

### LINKED ISSUE #879: Accordion: two digit numbers in lists not showing
When we have long lists inside accordions, the numbers with two digits only display the second digit.

Screenshot 2024-03-19 at 15 14 57

Examples:
https://iloorgstg.prod.acquia-sites.com/ines-green-jobs-portal
https://www-preview.ilo.org/node/644271/latest

justintemps:
I originally set out to fix #879. 

In short, when the list is ordered and in a container that has `overflow: hidden`, part of the numbers get lopped off because the marker is being rendered [outside the list](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position). Simply using `list-style-position: inside` makes the marker visible again, however it breaks the indentation from the second line onward.

I eventually ended up solving this using pseudo-elements, however, in the process I identified several other issues that made a general refactor of the styles the easiest way to fix this and other problems. Some of those included:

- Classnames didn't follow the modified BEM convention that we're using
- There were a lot of unnecessary classnames that were adding needless complication
- The unordered List was broken for RTL layouts

This fixes the issue in the List component itself, but I think the issue may have first arisen from a list that was rendered in a RichText component that was rendered inside an Accordion. If that's the case, then we'll need to carry over these styles to the RichText component, which may be the next step here.

This includes some changes to the Twig template that will also need to be carried over to the React component. I'll include that here as well.

**See Edits Below**


Shashika6:
Hey @justintemps had a look at this and it lgtm.

PR #905, opened 2024-04-02
GH-904: fixed the substitution of the prefix

### LINKED ISSUE #904: Twig: Fix the prefix substitution in the twig files outputted to the package
Currently the prefix in the twig files is being badly substituted while the package is being built (instead of being substituted with `ilo`, it is being substituted as `prefix: ilo`.) It should be fixed to substitute prefix as `ilo`.

bashlk:
Fixes #904 

The prefix in the twig templates was not being substituted correctly as the twig package was built. This PR fixes that and also cleans up the script that copies the files and the README.

### DIFF
--- a/.changeset/young-elephants-sin.md
+++ b/.changeset/young-elephants-sin.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": minor
+---
+
+Fixed the substitution of the prefix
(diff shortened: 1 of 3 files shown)


justintemps (approved):
Looks good to me. @ademarco ?

Issue #863, opened 2024-03-11
Twig/React: Navigation tab navigation doesn't work


justintemps:
It's not possible to tab through the options in the Nav component. Something is breaking tab navigation. This happens in both React and Twig.

https://twig.ui.ilo.org/iframe.html?args=&id=components-navigation--navigation&viewMode=story

If you start tabbing:

1. First you select the language selector (bad), you should get the link in the logo first
2. You can't tab through the options in the language selector (bad)
3. Then you get the More+ tab (bad), you should get the other options first from left to right (or right to left in RTL)
4. Then you get the Search tab. If you continue tabbing, instead of restarting the tab navigation it opens the search window, which is bad. Only activating the search button should reveal the tab and bring the search into focus.

Shashika6:
Update on this @justintemps making good progress so far will open up a PR for this tommorow.

PR #909, opened 2024-04-03
fix(react/twig) - Tab navigation improvements in navigation 

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/863

### Notes :- 

* Currently the tab navigation is not working as intended. 

### Issues fixed :- 

- [x] Language selector is selected first instead of the logo.
- [x] Can't tab through the options in the language selector.
- [x] When you get to the Search tab. If you continue tabbing, instead of restarting the tab navigation it opens the search window.

### Demo :- 

https://github.com/international-labour-organization/designsystem/assets/32934169/50379699-fbb7-4b8a-ae9f-3ea3fcbfd2ed

### Solution :- 

* Refactor CSS to use display instead of visibility


justintemps:
Hey @Shashika6 this works as expected on Chrome, but not on Firefox. Can you check?

Shashika6:
Will test it out in firefox with your settings as well. @justintemps

Shashika6:
Hey @justintemps had a look into this and finally figured out the issue on why this is happening. 

* I tested the preview in **windows with firefox** and everything is working as expected. 

* This is a known issue in **mac os with firefox**.

**Cause :** 

* By default the keyboard navigation is toggled off in mac os. 
* For the tab navigation to work in mac os with firefox we have to enable this toggle first from system settings. 
* After enabling the toggle the tab navigation should work as expected. 

How to enable keyboard navigation. 

System settings -> Keyboard -> Keyboard Navigation

justintemps:
Hey @Shashika6 you're right about tab navigation on Firefox, thanks for the tip!

As for this solution, you're right that it solves the problem, but in doing so it breaks the animations when the search windows and submenu scroll out from the bottom. We should be able to have both nice transitions as well as an accessible menu. 

Using `display: none` might well be part of the solution here, but I think you're going to have use some JavaScript to time it so that it's not added until after the menu transitions are finished.

justintemps:
@Shashika6 on closer inspection actually, I don't think we need `display: none` at all here, we can just dynamically set the tab index on the sub menu. I'm going to do this now.

justintemps:
@Shashika6 on closer inspection, setting tab index won't work because it's not an inherited property. So I went back to the first idea of dynamically adding `display: none` after the close transition was run and then removing it before the open transition. This was a bit of a hack, but I checked in Chrome, Safari and Firefox and it seems to work everywhere.

Shashika6:
Damn, this was a very tricky one good one @justintemps 🎉

justintemps:
Hey @Shashika6, this should be ready to check now, could you have a look? make sure to test in mobile as well please. I'll try to test this with bb8 after lunch.

Shashika6:
Works well in both firefox and chrome @justintemps should test this in mobile.

justintemps:
> Works well in both firefox and chrome @justintemps should test this in mobile.

Yea would you mind doing that?

Shashika6:
Navigation works as expected in mobile as well @justintemps

PR #921, opened 2024-04-05
Twig: Version 1 Release

justintemps:
## Changes from `@ilo-org/twig@0.16.0`

- Fix Accordion accessibility issues
- Fix design bug in pagination for RTL layouts
- Fix lighthouse accessibility audit error of tabs attributes not matching their roles
- Fixes some layout issues in List that caused parts of the list item numbers to get chopped off in ordered lists when they were rendered in containers with hidden overflow, as well as some incorrect rendering in RTL layouts.
- Refactor of how Twig outputs templates and bundles JavaScript for components. We have left the legacy implementation in place, for now, so this shouldn’t impact the website at all.
- Fixed broken styling in the Breadcrumb
- Fix Wingsuit definition making sure that they can be properly read by an actual Drupal site (you’re not using this, so it shouldn’t affect you)
Also, remove excessive use of "|raw" in templates, which was flagged a security issue.
- Field "id" was renamed to "elementId" on "tag" and "list" components. This was necessary since "id" is a reserved Drupal word. This also impacts the "tags" component, since it includes several "tag"s. We checked to make sure you’re not using any of these fields so this shouldn’t affect the website.
- Address accessibility issues with the heading order of cards by changing the default element used to render the title from an `<h3>` to `<p>`. Additionaly add an optional prop that allows for configuration of the title element if needed.


GGKapanadze:
There were no major defects detected with BB8 testing 

Just 2 minor UI flaws 

1. The edge of the language switcher is too sharp, noticeable during hover

https://github.com/international-labour-organization/designsystem/assets/36326203/48137027-dd87-4f13-bc08-2eefdf81e997

2. Extra spacing on info icon on mobile screen
image

GGKapanadze:
> There were no major defects detected with BB8 testing
> 
> Just 2 minor UI flaws
> 
> 1. The edge of the language switcher is too sharp, noticeable during hover
> 
>  Screen.Recording.2024-04-08.at.11.27.00.mov 
> 2. Extra spacing on info icon on mobile screen
>    image

The second issue seems to appear only on chrome devtools 
https://github.com/international-labour-organization/designsystem/assets/36326203/8f383b18-3cd2-48f5-8c2c-ad144b8455a3

PR #937, opened 2024-04-10
feat(twig-styles): accordion title always in one line

PauZoocha:
https://zoocha.atlassian.net/browse/ILO-253

This is to amend this requirement: 
`Text always shows in one line. Text truncates according to screen size.` using css and adding the <p> to the button label.

Screenshot 2024-04-10 at 16 06 49

### DIFF
--- a/.changeset/weak-ants-perform.md
+++ b/.changeset/weak-ants-perform.md
@@ -1,0 +1,6 @@
+---
+"@ilo-org/styles": patch
+"@ilo-org/twig": patch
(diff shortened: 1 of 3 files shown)


justintemps:
Hey @PauZoocha thanks for this but @inesdgomes came to the conclusion that are some situations where Accordion titles will need to be more than one line long. We've opted for a different solution where we just increase the amount of padding between the title and the edge of the Accordion Item.

Issue #935, opened 2024-04-10
Hyperlinks show odd underline on hover


inesdgomes:
When we hover long hyperlinks (especially those that show in two lines), the underline is meant to become thicker but this is not happening in the full line.
<img width="853" alt="Screenshot 2024-04-10 at 15 45 13" src="https://github.com/international-labour-organization/designsystem/assets/22772102/16790f0b-efee-4150-8492-76930de3a561">

Expand the 10th accordion on this page to see the issue: https://www-preview.ilo.org/resource/other/qas-business-and-child-labour#Why_does_ILO_not_consider_that_children_working_in_farms_as_part_of_their_education_ 

One more example: https://www-preview.ilo.org/resource/other/training-opportunities

inesdgomes:
@bashlk @GGKapanadze As Justin is on leave, is this something you could look into? Thanks!

Shashika6:
I can pick this one up @inesdgomes :)

justintemps:
Hey @inesdgomes here is the ticket where we're proposing increasing the Line height. Evidently right now it causes issues with the links which is why it doesn't show the proper underline.

inesdgomes:
@beatrizmartinmartins Could you look into this one please? @justintemps is proposing that we increase the line height to 1.7 in body text to avoid the issue with the underline. Could you please test this change in the designs and see if it works?

beatrizmartinmartins:
Hey @inesdgomes and @justintemps I just found the issue. It's a design and dev issue. 

While hovering the link on the live website, I found that we are using a background color on the hover. We should remove the background color in order for it to work. I just removed it on design as well. Just by doing this, we don't need to change any line-height (that will affect much more places).

![Image](https://github.com/international-labour-organization/designsystem/assets/96917981/cd9ed019-e72e-45ea-a808-c51646d85132)

inesdgomes:
Hi @beatrizmartinmartins Sounds like a good solution for 'enabled' hyperlinks in light backgrounds - but what happens with:
- 'enabled' hyperlinks with dark background? right now they have white background and blue text. 
- 'focus' hyperlinks on both light and dark backgrounds? they have a yellow background now.

beatrizmartinmartins:
We need to change the text-underline-offset to 0.25rem so we are able to make hove on the hyperlink without overlapping the next line 

[...]

PR #942, opened 2024-04-13
fix(style) - Increase line height on hyperlink to prevent thickness inconsistency

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/935

### Notes :- 

* Currently hyperlinks show an inconsistent underline on hover.

### Screenshot :- 

Before  :- 

Screenshot 2024-04-10 at 15 45 13

After :- 

Screenshot 2024-04-13 at 18 59 41

### Fix :- 

* Increase line height to have more spacing

### DIFF
--- a/.changeset/nice-ducks-design.md
+++ b/.changeset/nice-ducks-design.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Fix hyperlink underline showing thickness inconsistency on hover
--- a/packages/styles/scss/_mixins.scss
+++ b/packages/styles/scss/_mixins.scss
@@ -244,7 +244,7 @@
   color: map-get($color, "link", "text", "default");
   font-size: inherit;
   font-weight: inherit;
-  line-height: inherit;
+  line-height: 1.7;
   text-decoration: underline;
   text-underline-offset: px-to-rem(6px);
   text-decoration-thickness: px-to-rem(1.5px);


justintemps:
Hey @Shashika6 this does indeed fix the problem, but I don't think it's a good idea to only increase the line height for links. What about the rest of the text. If this is happening in the Rich Text component, it seems like we should increase the base line-height for the whole thing, no?

Shashika6:
Yes @justintemps I wanted to have a chat with you regarding this once you were back.

Shashika6:
Some context on the issue. 

This issue happens in every place where links are used and hovered on. This includes components like table as well. 

The reason why this happens is there isn't enough space for the underline hover thickness when there's content more than one line for the link.

Suggested solutions :- 

1. Increase the line-height where links as used. 
2. Reduce the size of the underline thickness for hover. 
3. Change the color on hover instead of increasing the thickness for hover. 

What do you think ? @bashlk @justintemps

bashlk:
Personally, I would lean towards reducing the underline thickness and changing the color to show emphasis on the link during hover. Changing the line height would force us to account for it in a lot of circumstances and I don't think it is worth it.

Thoughts @inesdgomes?

inesdgomes:
Hi @bashlk I agree, also not very keen on changing the line height as it may have an unforeseen impact on other components. @beatrizmartinmartins what do you think?

justintemps:
In the end, I was able to fix this just by removing the background from links in the RichText component.

PR #970, opened 2024-04-19
Template for CSS housekeeping milestone

justintemps:
This PR provides an example of the kind of refactoring/housekeeping I think we should do for all of the components, one by one, as a part of fulfilling this milestone.

For each component, I think we should:

- [ ] Replace explicit instances of `ilo` with the `$prefix` variable to enable theming later on
- [ ] Make sure classnames conform to modified BEM convention (`block--element__modifier`)
- [ ] Remove local variables that should instead use standard ones from the theme
- [ ] Explain any local variables that have to be used for special cases
- [ ] Remove unneeded rules
- [ ] Replace `map-get` invocations with the name of the css variable exported by the themes package
- [ ] Use LTR/RTL agnostic css (Example `padding-inline-start` instead of `padding-left`)
- [ ] Where necessary, break up the CSS for very big components into logical sub-components, that could potentially be exported from the Design System by themselves


packages/styles/scss/components/_accordion.scss
```diff
@@ -1,43 +1,44 @@
 @use "../tokens" as *;
 @use "../functions" as *;
 @import "../mixins";
+@import "../config";
 
-.ilo--accordion {
+.#{$prefix}--accordion {
```
justintemps, line 6: Change explicit ilo to use prefix variable

justintemps, line 6: Change specific ilo to use prefix variable

packages/styles/scss/components/_accordion.scss
```diff
@@ -1,43 +1,44 @@
...
+  // This provides a reference to the outerclass within nested declarations
+  $accordion: &;
```
justintemps, line 8: Use sensibly named references to refer to the outer class within nested declarations

packages/styles/scss/components/_accordion.scss
```diff
@@ -1,43 +1,44 @@
...
+
+  // Accordion contents can add lots of height so the transition timing can't
```
justintemps, line 10: Explain why local variables have to be used

packages/styles/scss/components/_accordion.scss
```diff
@@ -1,43 +1,44 @@
...
   &--button {
     display: flex;
-    align-items: center;
     flex-direction: row;
     justify-content: space-between;
-    text-align: start;
     width: 100%;
     margin: 0;
-    padding: spacing(4) 0;
+    padding-block: spacing(4);
```
justintemps, line 23: Use direction agonistic css

packages/styles/scss/components/_accordion.scss
```diff
@@ -1,43 +1,44 @@
...
-
     background-color: $color-ux-background-default;
     background-position: calc(100% - px-to-rem(6px)) center;
     background-repeat: no-repeat;
     background-size: px-to-rem(24px) px-to-rem(24px);
-
     border: none;
-    border-top: 2px solid map-get($color, "accordion", "border");
-
+    border-top: 2px solid $color-base-neutrals-lighter;
```
justintemps, line 30: Remove `map-get` invocations

packages/styles/scss/components/_accordion.scss
```diff
@@ -1,43 +1,44 @@
...
-    font-weight: map-get($type, "weights", "section");
-    @include font-styles("label-medium");
-
+    font-weight: $type-weights-section;
     transition: all $transition-timing ease-in-out;
+    @include font-styles("label-medium");
     @include dataurlicon("add", $color-ux-labels-actionable);
 
-    &--large {
+    &__large {
```
justintemps, line 39: Use correct BEM convention

Shashika6 (approved):

PR #973, opened 2024-04-19
Twig: fix boolean settings in patterns and components

ademarco:


justintemps:
Hey @ademarco thanks for doing this, did you make sure that all of the values you updated still work in storybook? I checked a few that didn't look like they were working. Also a few places where the filter got added to the twig, but the pattern file didn't change.

ademarco:
@justintemps I did test the manually when I changed them but I'll further test them now, that I have completed a first iteration of changes.

Shashika6:
@ademarco Reviewed the components individually. Most of the components are working as expected I found small issues in the ones below.

* The button component looks a bit off with the booleans. I think this can be tackled with the button cleanup that I'm working on
* The hero component tooltip icon is not displaying.
* The tooltip icon boolean is not working as expected. 

-----

*  Tags active state doesn't seem to function as expected as well but this issue exists in the twig currently so looks like we will have to create a seperate issue for this one @justintemps

ademarco:
Thank you @Shashika6  I've fixed the following:

- The button component looks a bit off with the booleans. I think this can be tackled with the button cleanup that I'm working on
- The hero component tooltip icon is not displaying.
- The tooltip icon boolean is not working as expected.

Please open a follow up for the active state, since this is present also in the current implementation.

justintemps:
Hey @ademarco what will happen to users who don't have the boolval filter configured?

ademarco:
@justintemps they need it loaded in their twig env, I have to add an entry to the readme, suggesting which library to integrate into their project. For ILO Base Theme users it will be transparent once this one gets in https://github.com/international-labour-organization/ilo_base_theme/issues/13

Issue #947, opened 2024-04-16
Implement test setup for twig components


bashlk:
Setup the npm scripts and GitHub workflow to run cypress locally and on GitHub Actions.

bashlk:
@justintemps Why are the tests also being run on Mac OS?

justintemps:
@bashlk because I use a Mac and I wanted to avoid people making PR that worked in Ubuntu but not on my local setup.

PR #984, opened 2024-04-23
test(twig): added local cypress setup

bashlk:
A part of #947.

This PR adds the basic setup for running tests against the twig base theme container locally.

Current issues
- The base theme copies index.css directly from the styles packages while not declaring the styles package as an explicit dependency. It should be changed to copy index.css from the twig package. Previously, the twig package was not exporting the correct CSS file with all the styles for the components. I fixed that in this PR.  (Raised https://github.com/international-labour-organization/ilo_base_theme/issues/11 to resolve)
- The base theme image has caching enabled and the cache needs to be manually cleared whenever the twig components are changed for those changes to be visible on the theme site. We should probably disable the cache in the image since the image will only be used for development purposes. (Raised https://github.com/international-labour-organization/ilo_base_theme/issues/10 to resolve)
- The twig package needs to be built manually for those changes to appear on the theme site. If this proves to be an annoyance, we can attempt to implement a "watching build" that continuously builds the changes made to the project. But this would require some development effort.

### DIFF
--- a/.changeset/selfish-sheep-battle.md
+++ b/.changeset/selfish-sheep-battle.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": patch
+---
+
+Added cypress test setup with theme container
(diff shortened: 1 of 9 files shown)


bashlk:
This PR is ready to merge.

Issue #971, opened 2024-04-19
Basic housekeeping for all of the component CSS


justintemps:
The CSS is a bit unruly in many of the components in the Design System. We need to clean up and organise the files so they're more consistent and easier to reason around. This should ultimately make the `styles` and `themes` packages more maintainable and ideally reduce the main css bundle size, albeit probably by only a little.

For each component, I think we should:

- Replace explicit instances of `ilo` with the `$prefix` variable to enable theming later on
- Make sure classnames conform to modified BEM convention (`block--element__modifier`)
- Remove local variables that should instead use standard ones from the theme
- Delete any theme values that are specific to only one component
- Explain any local variables that have to be used for special cases
- Remove any unneeded rules
- Replace `map-get` invocations with the name of the css variable exported by the themes package
- Use LTR/RTL agnostic css (Example `padding-inline-start` instead of `padding-left`)
- Where necessary, break up the CSS for very big components into logical sub-components, that could potentially be exported from the Design System by themselves

I've done this with the Accordion component and made a[ Pull Request](https://github.com/international-labour-organization/designsystem/pull/970) to show an example for what I have in mind.

## Stylesheets to tidy

- [x] `_accordion.scss`
- [x] `_breadcrumb.scss`
- [x] `_button.scss`
- [x] `_callout.scss`
- [ ] `_card.scss`
- [ ] `_cardgroup.scss`
- [ ] `_checkbox.scss`
- [ ] `_container.scss`
- [ ] `_contextmenu.scss`
- [ ] `_credit.scss`
- [ ] `_datacard.scss`
- [ ] `_datepicker.scss`
- [ ] `_detailcard.scss`
- [ ] `_dropdown.scss`
- [ ] `_empty.scss`
- [ ] `_factlistcard.scss`
- [ ] `_featurecard.scss`
- [ ] `_fieldset.scss`
- [ ] `_file-upload.scss`
- [ ] `_footer.scss`
- [ ] `_form.scss`
- [ ] `_formcontrol.scss`
- [ ] `_heading.scss`
- [ ] `_hero.scss`
- [ ] `_herocard.scss`
- [ ] `_icon.scss`
- [ ] `_image.scss`
- [ ] `_input.scss`
- [ ] `_link.scss`
- [ ] `_linklist.scss`
- [ ] `_list.scss`
- [ ] `_loading.scss`
- [ ] `_logo.scss`
- [ ] `_logogrid.scss`
- [ ] `_modal.scss`
[...]

PR #994, opened 2024-04-28
fix(style) - housekeeping for button css

Shashika6:
This PR provides refactoring/housekeeping of the **button** component.

- [x] Replace explicit instances of `ilo` with the `$prefix` variable to enable theming later on
- [x] Make sure classnames conform to modified BEM convention (`block--element__modifier`)
- [x] Remove local variables that should instead use standard ones from the theme
- [x] Explain any local variables that have to be used for special cases
- [x] Remove unneeded rules
- [x] Replace `map-get` invocations with the name of the css variable exported by the themes package
- [x] Use LTR/RTL agnostic css (Example `padding-inline-start` instead of `padding-left`)
- [x] Where necessary, break up the CSS for very big components into logical sub-components, that could potentially be exported from the Design System by themselves

Related to #971 

**Design bugs fixed :-** 

**Align the small button icon and label**

Before :- 
Screenshot 2024-05-06 at 16 57 33

After :- 
Screenshot 2024-05-06 at 17 00 28

**Fix icon only component in twig**

Before :- 
Screenshot 2024-05-06 at 17 02 44

After :- 
Screenshot 2024-05-06 at 17 02 54

### DIFF
--- a/packages/styles/scss/components/_button.scss
+++ b/packages/styles/scss/components/_button.scss
@@ -2,23 +2,35 @@
 @use "../functions" as *;
 @import "../animations";
 @import "../mixins";
+@import "../config";
 
-.ilo--button {
+.#{$prefix}--button {
+  --ilo-button-label-weight: 700;
+  --ilo-button-labels-actionable-color: var(--ilo-color-blue-dark);
+  --ilo-button-labels-active-color: var(--ilo-color-blue-light);
+  --ilo-button-labels-alert-color: var(--ilo-color-white);
+  --ilo-button-labels-hover-color: var(--ilo-color-blue);
+
+  // This provides a reference to the outerclass within nested declarations
+  $button: &;
   display: inline-block;
   font-family: var(--ilo-fonts-display);
-  font-weight: map-get($type, "weights", "label");
+  font-weight: var(--ilo-button-label-weight);
   padding: 0;
-  @include borderradius("button");
+  border-radius: var(--ilo-border-md);
+  @include globaltransition("color, background-color, border-color");
 
   .link__label,
   .button__label {
     display: inline-block;
+    position: relative;
+    top: px-to-rem(2px);
   }
 
-  &.icon {
+  &--icon {
     position: relative;
 
-    .ilo--icon {
+    .#{$prefix}--icon {
       height: 100%;
       max-height: px-to-rem(32px);
       max-width: px-to-rem(32px);
@@ -26,25 +38,25 @@
       width: 100%;
     }
 
-    &.icon__position--left {
+    &--position__left {
       .link__label,
       .button__label {
-        padding-left: spacing(12);
+        padding-inline-start: spacing(6);
       }
 
-      .ilo--icon {
+      .#{$prefix}--icon {
         left: px-to-rem(13px);
         top: px-to-rem(7px);
       }
     }
 
-    &.icon__position--right {
+    &--position__right {
       .link__label,
       .button__label {
-        padding-right: spacing(12);
+        padding-inline-end: spacing(6);
       }
 
-      .ilo--icon {
+      .#{$prefix}--icon {
         right: px-to-rem(13px);
         top: px-to-rem(7px);
       }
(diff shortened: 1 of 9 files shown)


Shashika6:
@justintemps I noticed that there is a border radius token just used for the button. We should get rid of this one right? 

* The reason why I didn't delete it yet is because its been used in fileupload as well. Should we come back to this on file upload ?

packages/themes/tokens/radius/button.json

justintemps (changes requested):
Hi @Shashika6 think you're still missing some things here

packages/styles/scss/components/_button.scss
```diff
@@ -2,13 +2,18 @@
...
+.#{$prefix}--button {
+  // This provides a reference to the outerclass within nested declarations
+  $button: &;
   display: inline-block;
   font-family: $fonts-display;
-  font-weight: map-get($type, "weights", "label");
+  font-weight: $type-weights-label;
   padding: 0;
-  @include borderradius("button");
+  border-radius: $radius-button-top-left $radius-button-top-right
```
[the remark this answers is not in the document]
Shashika6: This is used in like 6 components so I think it should be fine right what do you think @GGKapanadze

justintemps:
@GGKapanadze could you give this a look?

packages/styles/scss/components/_button.scss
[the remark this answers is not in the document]
GGKapanadze: Of course, if it's button-specific we should define it on the button level but if it's general we can move it to the foundation after #999

GGKapanadze:
@justintemps @Shashika6 `data-js-processed` selector is removed and replaced with twig component attributes

GGKapanadze (commented):
Everything looks good, but we can hold it to integrate with #999

Shashika6:
@justintemps [#1084](https://github.com/international-labour-organization/designsystem/issues/1084) Is done in this as well. Will update the changesets along with the variables and let you know.

packages/styles/scss/theme/_foundation.scss
```diff
@@ -26,6 +26,8 @@
   --ilo-fonts-monospace: monospace;
   --ilo-line-height: 1.46;
 
+  --ilo-types-weight-label: 700;
```
bashlk, line 29: @GGKapanadze Shouldn't this be --ilo-font-weight-bold? We are moving away from context based naming right?

packages/styles/scss/components/_button.scss
```diff
@@ -57,269 +63,204 @@
...
+    border: var(--ilo-border-base) var(--ilo-color-borders-focus) solid;
+    box-shadow: 4px 4px 0 1px var(--ilo-color-borders-focus) inset,
+      -4px -4px 0 1px var(--ilo-color-borders-focus) inset;
+    color: var(--ilo-color-labels-focus);
+    outline: none;
+    @include globaltransition("color, background-color, border-color");
 
-    &--large.icon--only {
-      transform: translateY(-6px);
+    &.ilo--button--small {
```
bashlk, line 239: This still has a hardcoded prefix

packages/styles/scss/theme/_foundation.scss
```diff
@@ -61,6 +63,36 @@
 
   --ilo-color-purple: rgba(150, 10, 85, 1);
 
+  --ilo-color-notification-icon: var(--ilo-color-white);
```
bashlk, line 66: If it's the same color, we should use that token directly instead of redeclaring it like this right? @GGKapanadze

packages/styles/scss/theme/_foundation.scss
GGKapanadze: This should be declared as a component variable

packages/styles/scss/theme/_foundation.scss
```diff
@@ -61,6 +63,36 @@
...
+  --ilo-color-notification-type-error: var(--ilo-color-red);
+  --ilo-color-notification-type-info: var(--ilo-color-blue);
+  --ilo-color-notification-type-success: var(--ilo-color-green);
+  --ilo-color-notification-type-warning: var(--ilo-color-yellow);
+  --ilo-grey-light: rgba(237, 240, 242, 1);
+  --ilo-grey-ui: rgba(184, 196, 204, 1);
```
GGKapanadze, line 72: if it's colors follow the color convension, if it's only used in one component make it component specific

packages/styles/scss/theme/_foundation.scss
```diff
@@ -61,6 +63,36 @@
...
+
+  --ilo-color-background-hover: var(--ilo-color-blue-lighter);
+  --ilo-color-background-focus: var(--ilo-color-white);
+  --ilo-color-background-alert: var(--ilo-color-red-dark);
+
+  --ilo-color-labels-actionable: var(--ilo-color-blue-dark);
+  --ilo-color-labels-active: var(--ilo-color-blue-light);
+  --ilo-color-labels-alert: var(--ilo-color-white);
+  --ilo-color-labels-hover: var(--ilo-color-blue);
+  --ilo-color-labels-focus: var(--ilo-color-blue-dark);
```
GGKapanadze, line 94: I think this also is a component level variables

GGKapanadze (changes requested):

packages/styles/scss/theme/_foundation.scss
Shashika6: This color is used in many components so it should be in the foundation :)

GGKapanadze: Than we should name it accordingly, like other colors

Shashika6: Cool will do @GGKapanadze

Shashika6:
Refactored based on the feedback @GGKapanadze removed a few unused variables from the foundation and moved variables to component level where needed

packages/styles/scss/theme/_foundation.scss
```diff
@@ -90,4 +101,6 @@
   --ilo-border-sm: calc(#{px-to-rem(1.5)} * var(--ilo-scale));
   --ilo-border-md: calc(#{px-to-rem(2)} * var(--ilo-scale));
   --ilo-border-lg: calc(#{px-to-rem(4)} * var(--ilo-scale));
+
+  --ilo-border-base: var(--ilo-border-md);
```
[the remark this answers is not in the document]
Shashika6: Sure @GGKapanadze

Shashika6:
@GGKapanadze This is good to go now right Cc :- @justintemps

justintemps:
Hey @Shashika6 I merged #1000 and now this has some merge conflicts in the Foundations file. Could you resolve those and rebase this on top of `develop`?

PR #1000, opened 2024-04-29
fix(style) - housekeeping for callout css

Shashika6:
This PR provides refactoring/housekeeping of the **callout** component.

- [x] Replace explicit instances of `ilo` with the `$prefix` variable to enable theming later on
- [x] Make sure classnames conform to modified BEM convention (`block--element__modifier`)
- [x] Remove local variables that should instead use standard ones from the theme
- [x] Explain any local variables that have to be used for special cases
- [x] Remove unneeded rules
- [x] Replace `map-get` invocations with the name of the css variable exported by the themes package
- [x] Use LTR/RTL agnostic css (Example `padding-inline-start` instead of `padding-left`)
- [x] Where necessary, break up the CSS for very big components into logical sub-components, that could potentially be exported from the Design System by themselves

Related to #971


packages/styles/scss/components/_callout.scss
```diff
@@ -16,42 +17,45 @@
     max-height: 300px;
     overflow: hidden;
 
-    #{$self}--toggle--icon {
+    #{$callout}--toggle__icon {
```
justintemps, line 20: This doesn't follow the standard. Should be something like `#{$callout}__toggle-icon`

Shashika6:
Did the changes @justintemps

justintemps:
@GGKapanadze could you give this a look?

Shashika6:
@justintemps @GGKapanadze Can you have a look at this I applied css variables to this as well.

Shashika6:
The dataurlicon mixin needs to be modified to factor in css variables. Which is available in a lot of components will be creating a seperate issue for that.

packages/styles/scss/theme/_foundation.scss
```diff
@@ -61,6 +61,12 @@
 
   --ilo-color-purple: rgba(150, 10, 85, 1);
 
+  --ilo-color-ux-notification-icon: var(--ilo-color-white);
+  --ilo-color-ux-notification-type-error: var(--ilo-color-red);
+  --ilo-color-ux-notification-type-info: var(--ilo-color-blue);
+  --ilo-color-ux-notification-type-success: var(--ilo-color-green);
+  --ilo-color-ux-notification-type-warning: var(--ilo-color-yellow);
```
GGKapanadze, line 68: I think we can get rid of ux prefix

GGKapanadze (changes requested):

packages/styles/scss/theme/_foundation.scss
GGKapanadze: and also, `icon` doesn't need to have color definition, just reference the color directly for that one

Shashika6: Sure @GGKapanadze

GGKapanadze (approved):

justintemps (approved):

PR #1002, opened 2024-05-02
React: Align Breadcrumb with Twig

justintemps:
This provides a complete refactor of the Breadcrumb and fixes several accessibility issues in the process.

### DIFF
--- a/packages/react/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/packages/react/src/components/Breadcrumb/Breadcrumb.tsx
@@ -1,7 +1,14 @@
 /* eslint-disable jsx-a11y/anchor-has-content */
 /* This is a good rule in general, but disabling here because the link has an aria-label */
 
-import { FC, useState } from "react";
+import {
+  FC,
+  useState,
+  useEffect,
+  useLayoutEffect,
+  useCallback,
+  useRef,
+} from "react";
 import useGlobalSettings from "../../hooks/useGlobalSettings";
 import classnames from "classnames";
 import { BreadcrumbProps, BreadcrumbItemProps } from "./Breadcrumb.props";
@@ -14,8 +21,6 @@ const BreadcrumbItem: FC<BreadcrumbItemProps> = ({
   labelShown = true,
 }) => {
   const { prefix } = useGlobalSettings();
-
-  const [contextMenuOpen, setContextMenuOpen] = useState(false);
 
   const baseClass = `${prefix}--breadcrumb--item`;
   const linkClass = `${prefix}--breadcrumb--link`;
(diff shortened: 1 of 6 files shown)


justintemps:
@Shashika6 and @GGKapanadze, has one of you guys had a chance to look at this?

Shashika6:
No @justintemps , @GGKapanadze was having a look at this one

justintemps:
Hey @GGKapanadze any ideas?

packages/react/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,80 +1,288 @@
...
+    }
+  }, [isMenuCollapsed, isCtxMenuOpen]);
 
-  const showContext = links.length > 4 || showContextMenu ? true : false;
+  // handle tab navigation when the context menu is open
+  useEffect(() => {
+    function handleKeyDown(event: KeyboardEvent) {
+      if (!isCtxMenuOpen || !isMenuCollapsed) {
+        return;
+      }
```
GGKapanadze, line 181: Let's check if refs are initialized to be more consistent like we are doing in `useLayoutEffect`

packages/react/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,80 +1,288 @@
...
+
+    observer.observe(document.body);
+
+    return () => {
+      observer.disconnect();
+    };
+  }, [collapseMenu]);
+
+  // Position the Context Menu right below the button
+  useLayoutEffect(() => {
```
GGKapanadze, line 134: I have this issue with the breadcrumbs context placment

https://github.com/international-labour-organization/designsystem/assets/36326203/d39e51c0-6d22-4025-8b09-928826da1622

Testewed with:
- Chrome v124
- Firefox v125
- Safari

GGKapanadze (changes requested):

packages/react/src/components/Breadcrumb/Breadcrumb.tsx
justintemps: Hey @GGKapanadze, it turns out this was being caused by the Canvas wrapper that appears in the Docs page. To fix it, I'm just rendering the component in Storybook outside of a Canvas box and then showing the code beneath it using one of the native Storybook blocks.

The issue is that the Canvas container has a `transform` property which creates its own [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block). This in turn breaks the positioning of the Context Menu, which is expecting to calculate its position from the viewport and instead ends up calculating from the Canvas box.

Because I think it's very unlikely that a Breadcrumb would ever need to be rendered in a container with a transform set on it, I think we can let this go. But I did point out in the docs that that could have some undesirable consequences.

Issue #999, opened 2024-04-29
ILO Design System Theme Proposal


GGKapanadze:
# Context

The proposal idea comes after discussing the #896. 

First of all, let's define how the current implementation works. The mono-repo has a package named `@ilo-org/theme`, with a simple idea to define an abstract variable system that will be a source of truth for the `@ilo-org/styles` package. It uses JSON files to define the configuration, something like this: 

```json
  "spacing": {
    "base": { "value": "8px" },
    "root": { "value": "4px" } 
 }
```
After the definition, there is a  [built](https://github.com/international-labour-organization/designsystem/blob/develop/packages/themes/build.js) step involved(it uses `-` for the variable concatenation. ) that will produce the `SCSS` variables from the configuration that are directly accessible

```scss
{
  .example {
    margin-top: $spacing-base; 
  }
}
```

## Problem
Now, let's look at a more complex, real-life use case. For example, look at the `Hero` component and imagine we want to tweak its style. We will start from the `_hero.scss` file and we will notice the `$spacing-hero-card-corner-cut-height-lg` variable. If we want to modify the actual value. First, we will try to do the `go to definition` aka (`cmd` + click). We will end up inside the `_tokens.scss` file, which is a built file and **should not be modified**. Next, we'll try to use the global search, but none of them appears to be a JSON file 

![image](https://github.com/international-labour-organization/designsystem/assets/36326203/e02532fd-6f6f-4bec-a092-5b676cb31292)

To really modify the value, we should look at the variable structure, and we'll end up inside `spacing/hero.json`, and here it is 

```json
{
  "spacing": {
    "hero-card": {
      "corner-cut": {
        "width": {
          "sm": { "value": "72px" },
          "lg": { "value": "116px" }
        },
        "height": {
          "sm": { "value": "40px" },
          "lg": { "value": "64px" }
        }
      }
    }
  }
}
```

Despite a bad developer experience, this approach also introduced the dead code elimination problem. There is no easy way to detect unused tokens and purge them. This may evolve into a bigger maintenance problem that will take a lot of work to debug. You may also notice the second issue, as mentioned above: the `-` is used for variable concatenation for JSON nested values, but the `-` is also utilized for word separation. So unexpected collisions may happen:

```json
{
  "spacing": {
    "herox": {
      "card": {
        "padding": {
          "value": "11px"
        }
      }
    },
    "herox-card": {
      "padding": {
        "value": "18px"
      }
    }
  }
}
```
[...]

Issue #896, opened 2024-03-27
Styles: add default typography styles


justintemps:
We need to align the default typography applied to text outside of individual components to the specifications in this Foundations file on Figma: 

https://www.figma.com/file/qr8jQnxqJkRJFQnVC5PmrB/ILO-Foundations?type=design&node-id=6458-12248&mode=design&t=9jfxM7DV4iHAjayC-0

Right now, it seems to me like we don't apply any default typography at all. The `packages/styles/scss/_typography.scss` file just has some mixins borrowed from [Carbon](https://carbondesignsystem.com/). It's not clear to me where or how those are even being used.
 
First, we need to devise a strategy for applying typography styles in components. It's really inconsistent right now, in some cases we're getting values from the themes package, in others the styles are hard coded. We can apply the strategy gradually, but we need to agree on what it should be.

Then we need to use this strategy to apply the defaults described in the Foundations file.

GGKapanadze:
Reference for design Team @justintemps  @inesdgomes @beatrizmartinmartins 

- [Google Material Design](https://m2.material.io/design/typography/the-type-system.html#applying-the-type-scale)
- [Elastic UI](https://www.figma.com/file/oMt3fOymNwx87YsO9kGZhs/Elastic-UI-(Community)?type=design&node-id=10735-39974&mode=design&t=n6LUm1IVFdYJzkM6-0) 
- [Ant Design](https://www.figma.com/file/zcvLTQJKRlSSNXrCU9FSQr/Ant-Design-Open-Source-(Community)?type=design&node-id=473-956&mode=design&t=K0YgboqlB6iDU3H8-0)
- [IBM Carbon](https://carbondesignsystem.com/elements/typography/type-sets)

beatrizmartinmartins:
@GGKapanadze what do you need from our side in order to solve this issue? Should we focus on documentation about foundations? I know there's some issues on components as well, but the previous budget was not meant to change substancially any component, that's why we didn't included it in the scope. @inesdgomes what do you think we should do as well?

inesdgomes:
@beatrizmartinmartins Could you please review this comment: https://github.com/international-labour-organization/designsystem/issues/896#issuecomment-2083106776

beatrizmartinmartins:
[...]

PR #1030, opened 2024-05-21
ILO Theme Foundation

GGKapanadze:
## Foundation 

The PR introduces the foundation file mentioned in #999. The foundation file has 2 primary sources, the first one is the **Figma** foundation file, and the second one is the old token system or actual source code. There is also a third-ish one I named it **Mixed**, part from source code and part from Figma components

Foundation file 
✅ - Implemented
🚧 - Partially
⛔ - Not Implemented

| Scope | Status | Soure | 
|--------|--------|------|
| Colors | ✅ | Figma | 
| Typography | 🚧 |  Mixed | 
| Sizing | ✅ |  Figma | 
| Spacing | ✅ | Figma | 
| Breakpoints | ✅ | Source Code | 
| Transitions | ✅ |  Source Code | 
| Borders | ✅ |  Source Code | 

### Typography
 
Typograph is marked as not implemented because we need a Figma update for #896, but it has a main port of the current token system

### Mixins

Two widely used mixins were refactored
-  `spacing` mixin now is fully powered by a new foundation
-  `breakpoint` mixin was a bit problematic because referencing CSS variables as a media query value isn't possible without `js` interaction so I have used SCSS variables to define the breakpoints(for mixing usage) and also composed `CSS` variables from those, `SCSS`version of variables aren't directly referenced by any property it's only used for breakpoint mixin. 

### Found Figma Problems
We have 2 green colors, so I made the second one, `green-light`
image

### Prefixes

All foundation variables are prefixed as proposed in #999. Prefixes are static for now, but they can be implemented with concatenation if we really want to make it "dynamic".  I still suggest to keep it static for now because it's better for visual feedback, referencing and general readability, after we agree about "composability" of the designsystem we can always do plain old find and replace for `--ilo`

### DIFF
--- a/.changeset/chilly-llamas-grin.md
+++ b/.changeset/chilly-llamas-grin.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": minor
+---
+
+Add foundation file for the new theming
--- a/packages/styles/scss/_functions.scss
+++ b/packages/styles/scss/_functions.scss
@@ -57,11 +57,13 @@
 // =========================================
 
 @function spacing($size) {
-  $scaled: $size * $spacing-root;
+  $scaled: calc(#{$size} * var(--ilo-spacing-base));
+
   @if ($size % 1 != 0) {
-    $scaled: round($size) * $spacing-root;
+    $scaled: calc(#{round($size)} * var(--ilo-spacing-base));
   }
-  @return px-to-rem($scaled);
+
+  @return $scaled;
 }
 
 // =========================================
--- a/packages/styles/scss/_mixins.scss
+++ b/packages/styles/scss/_mixins.scss
@@ -3,6 +3,7 @@
 //*------------------------------------*/
 @use "tokens" as *;
 @use "functions" as *;
+@use "theme/foundation" as *;
 
 // ======================================
 // Output font styles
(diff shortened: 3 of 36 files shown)


justintemps:
Hey @GGKapanadze I love the direction here, this all looks fine, I'm still undecided about the prefixes. However, I was noticing that the main css bundle has gone from 271kb in the current release to 421kb in this one. I didn't have the chance to dig in and figure out why that was happening, do you have any ideas?

justintemps:
Hey @GGKapanadze actually I had a look and it seems like the root variables are all getting prepended to each class for some reason

GGKapanadze:
I'll try to fix it, I think it's about import resolution

justintemps:
Hey @GGKapanadze before I look at this again, did you have a second to look at the import resolution?

GGKapanadze:
@justintemps The issue was caused by the `scss` variable import inside the mixin, which importing the full `:root` definition for every file that was using the `mixins.scss`

> Old
Alacritty 2024-06-13 00 34 03

> New
Alacritty 2024-06-13 00 33 21

justintemps:
Hey @GGKapanadze , I was trying to test this out in bb8, but then I hit this error: https://github.com/international-labour-organization/designsystem-site-builder/issues/6

Could you have a look?

Shashika6:
Fixed the e2e test for tooltip. :')

Cc :- @justintemps @GGKapanadze

PR #1031, opened 2024-05-21
fix(style) - increase underline offset

Shashika6:
This is a draft PR to test the underline offset

### DIFF
--- a/packages/styles/scss/_mixins.scss
+++ b/packages/styles/scss/_mixins.scss
@@ -185,7 +185,7 @@
   font-weight: inherit;
   line-height: inherit;
   text-decoration: underline;
-  text-underline-offset: px-to-rem(6.2px);
+  text-underline-offset: px-to-rem(7px);
   text-decoration-thickness: px-to-rem(1.5px);
   padding-bottom: spacing(1);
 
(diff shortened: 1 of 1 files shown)


justintemps:
Hey @Shashika6 this still doesn't work predictably, see below

https://github.com/international-labour-organization/designsystem/assets/12401179/6a798f90-6abf-4df6-afb6-6b82f72e3ee5

Shashika6:
Can you check this now @justintemps

justintemps:
Still no good @Shashika6, the background color applied on background is cover up the underline

image

Shashika6:
Hey @justintemps I checked out with multiple values but there is no option that satisfies both the  offset issue and characters like g breaking the underline. 

A possible solution would be to increase the line-height what do you think ?

justintemps:
hey @Shashika6, I think the problem is the background color added to the link. What happens if we remove that in the RichText?

Shashika6:
Tried removing the background color in richtext @justintemps It doesn't seem to have any effect.

justintemps:
@Shashika6 this is all you had to do: https://github.com/international-labour-organization/designsystem/pull/1048/files

PR #1035, opened 2024-05-22
feat(styles): Breadcrumbs padding left to the inner wrapper

PauZoocha:
https://zoocha.atlassian.net/browse/ILOSD-37
This is to add padding left to the breadcrumbs inner wrapper for the breadcrumbs to be aligned on the site with the header.
eg. https://github.com/PauZoocha/designsystem/blob/develop/packages/styles/scss/components/_navigation.scss#L211

### DIFF
--- a/.changeset/cuddly-buses-sniff.md
+++ b/.changeset/cuddly-buses-sniff.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Add padding left to the breadcrumbs inner wrapper as per the header.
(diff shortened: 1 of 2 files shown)


justintemps:
Hey @PauZoocha thanks a lot for the PR, I totally get what you're trying to accomplish here, but I don't think we can merge this. I think we're having this issue because the Breadcrumbs are implemented in the website in a way that's not how they were intended to be used in the DS. I don't think we can change the internals of the component to accommodate non-standard use.

For the time being, you can accomplish what you're trying to do simply by setting the `--breadcrumb-padding` css variable on one of the Breadcrumbs parents.

This is how we manage the the Breadcrumbs positioning in the Hero component (where the Breadcrumb is intended to be used as a sub component). 

See here for reference. https://github.com/international-labour-organization/designsystem/blob/60609c8ae5e0098bc709540fb19ea1e4e6e20b68/packages/styles/scss/components/_breadcrumb.scss#L16

Issue #979, opened 2024-04-22
Twig: Improve Getting Started Page


justintemps:
The Twig "Getting Started" Page essentially needs to be rewritten. I'd recommend the following structure.

## Introduction

What is the Design System for Twig and when should it be used?

## Usage with the ILO Base Theme

This should explain how to incorporate the Design System for Twig using the [ILO Base Theme](https://github.com/international-labour-organization/ilo_base_theme). The project README here is a good place to start drafting this.

## Usage without the ILO Base Theme

Instructions on importing components directly from npm package.

ademarco:
@justintemps as discussed I've only kept the Drupal-related part for now. Please feel free to reorganize it as you see fit and/or merge it.

ademarco:
Reopening, as we still need to add the generic php section.

ademarco:
@justintemps I've created a follow up of this ticket here #1086. IMO we can merge this ticket's PR and close it.

justintemps:
Closed by #1065

PR #1065, opened 2024-06-17
Twig: Improve Getting Started Page

ademarco:
#979


justintemps:
We also wanted to add a page for use with generic PHP applications that might not necessarily be Drupal.

justintemps:
Hey @ademarco I restructured the docs pages a bit, see what you think. We can add instructions for a generic PHP project in packages/twig/apps/storybook/patterns/php.stories.mdx

Issue #1046, opened 2024-06-10
Twig: Upgrade to Wingsuit 2


justintemps:
This is a stub for upgrading to Wingsuit 2.

Here is a link to the upgrade guide: https://github.com/wingsuit-designsystem/wingsuit/blob/release-2-0/docs/src/app/(docs)/guides/ws2-upgrade/index.mdx

justintemps:
Hey @bashlk and @GGKapanadze, I'm beginning to think that upgrading to Wingsuit 2 is a dead end and that we should instead find a simpler way of rendering our Twig templates in Storybook.

I've been trying to upgrade to the beta version of Wingsuit 2 and it's been really hard going. Also, I feel like it's too much of an abstraction on top of Storybook, which is complicated enough as it is. 

I've prepared a proof of concept that uses a simple Vite plugin [vite-plugin-twig-drupal](https://github.com/larowlan/vite-plugin-twig-drupal) to render the templates. It does a good job of handling the Twig files and can even manage `@includes`.

However, it doesn't handle scripts, which gets a little tricky when components have sub components that have their own JS behaviors. We can either live with that or try to tweak the plugin workflow so that it compiles scripts as well as Twig templates. Anyway, give it a look and see what you think: https://github.com/justintemps/twig-vite-poc

bashlk:
@justintemps Why don't we stop using storybook altogether for the twig components?

Since we are making all the twig components follow the UI patterns spec for the base theme, we could also use the UI patterns environment to play around with the components themselves (See https://www.drupal.org/project/ui_patterns - the base theme docker image already presents this for us) This way, we would be using the same environment during development as a Drupal developer, which would also reduce the chance of us running into the Drupal specific issues that we have had in the past. 

We could still use storybook for the React version of the components so in summary, we would use the base theme docker image to develop the twig components and storybook to develop the React components. We could introduce more variants of the components so that we can preview more states of it and we can also use these for the tests.

The biggest concern I have with vita-plugin-twig-drupal is that there is no community behind it yet, it is mostly one developer working on it. It is the same developer in fact who created [twig-testing-library](https://github.com/larowlan/twig-testing-library) which hasn't been updated in 3 years. So it is likely that this plugin would also not be actively developed in the future.

justintemps:
[...]

Issue #1038, opened 2024-05-28
Core: upgrade to Node 20


justintemps:
Related to #1037 and blocked by #1068

We need to move the whole project onto Node 20.

PR #1066, opened 2024-06-18
Storybook Alternative

GGKapanadze:
# Storybook Alternative

This PR is a concept for #1046 and #1038. We are using `wingsuit-designsystem/wingsuit@1`, which isn't working for node `>16`. The second version of Wingsuit is in beta and has no valuable documentation or migration guide. So, we still need the storybook for Twig, but we also need to move from `node:16` and old storybook

Let's quickly list the main features that we want from Twig Storybook

- [ ] Resolve twig files
- [ ] Resolve components corresponding `JS` bundle
- [ ] Resolve nested components and their `JS` bundles 
- [ ] Drupal behavior setup
- [ ] Support `UI:Patterns` 

This PR tries to introduce the storybook that supports those features.

---

### 0. Base
The current branch uses `node:20`, and the `html` storybook, inside `twig-new` package 

### 1. Resolve twig files

For parsing and resolving the Twig files, it uses `larowlan/vite-plugin-twig-drupal`, so that's the first checkbox 

- [x] Resolve twig files

### 2. Resolving JS bundle

Resolving JS bundle for specific twig component is introduced as a patch for `larowlan/vite-plugin-twig-drupal` that injects the `behavior.js` file if it's presented with component

So this code is a valid import, that exports a `JS` function 
```js
import Button from "../components/button/button.twig";
```

- [x] Resolve components corresponding `JS` bundle

### 3. Resolve nested components and their `JS` bundles 

The `patch` is implemented in a way that JS becomes part of the actual twig component so there it works recursively for any `twig` includes 

```js
import Button from "../components/button/button.twig";
```

The code above resolves the component's `JS`, no additional import is required

- [x] Resolve nested components and their `JS` bundles 

### 4. Drupal behavior setup

The storybook wraps each story with `Drupal` globally, for behaviors to work correctly 

### 5. Support `UI:Patterns`

This is the last dragon to defeat; in order to support `pattern.yml` files, we first need to parse the `yml` definition file. For that, the implementation uses `Modyfi/vite-plugin-yaml.` So importing the `yml` file will resolve into `JS` object 

```ts
import Button from "../components/button/button.twig";
import ButtonPatterns from "../components/button/button.pattern.yml";
```

The next step is to generate a story based on `pattern` definition. This is done by a custom `compose` function that outputs both `Docs` and `Primary` stories. 

```ts
import Button from "../components/button/button.twig";
import ButtonPatterns from "../components/button/button.pattern.yml";

import { compose } from "../src/bridge/index";

const { root, Default } = compose(Button, ButtonPatterns);

const Meta = {
  title: "Components/User Interfaces/Button",
  tags: ["autodocs"],
  ...root,
};

export default Meta;
export { Default };
```

The story `title` and `tags` still have to be static, I tried to create a transform plugin for storybook stories to have the smoothest experience, but they are parsed outside of `vite` :/ , and I can't tweak exports.

---

## Here is a demo

https://github.com/international-labour-organization/designsystem/assets/36326203/7561de07-83ad-4880-b36f-e9da5aa23912

### Roadmap

- [ ]  explore `UI:Patterns` schema to support all definitions(`variants`, `nesting`,  etc)
- [ ]  came up with the migration strategy 
- [ ]  support CSS hot reloading 

We can use this PR to discuss the feature of `twig` package 

### Testing

- `clone` the branch
-  install `node:20`
-  enter inside `packages/twig-new`
- run `pnpm storybook`


packages/twig-new/src/bridge/index.js
```diff
@@ -0,0 +1,91 @@
+import { schema } from "./schema";
```
justintemps, line 1: Why `src/bridge`? It's a utility function, but it's not getting exported. May be `lib/compose` or something?

packages/twig-new/src/bridge/schema.js
```diff
@@ -0,0 +1,33 @@
+import { z } from "zod";
```
justintemps, line 1: Great idea.

packages/twig-new/stories/Button.stories.js
```diff
@@ -0,0 +1,15 @@
+import Button from "../components/button/button.twig";
```
justintemps, line 1: I like this

patches/vite-plugin-twig-drupal@1.3.0.patch
```diff
@@ -0,0 +1,70 @@
+diff --git a/src/index.js b/src/index.js
```
justintemps, line 1: Very cool! 

I thought about this for a long time last week and decided that it wasn't worth trying to bundle the JS together together with the Twig function that gets exported from the vite plugin. In the first place, Wingsuit doesn't do it, you have to explicitly import the JS for sub-components and in the second, it's unexpected, you wouldn't assume that a plugin for twig components would go looking for JS files too.

But I really like this implementation, it's straightforward and makes for nice story files.

I especially like using a the schema to check the validity of pattern files, that's a great touch.

In terms of patching the original vite-twig plugin, that's cool for now, but I think we should either:

1. Make a PR to support this (may be as an option) to the current library
2. Fork it and publish separately

justintemps:
@GGKapanadze 😬

image

PR #1089, opened 2024-06-28
Fix(styles/themes): Chinese fonts not rendering correctly

### LINKED ISSUE #986: Fonts: Chinese text mixes Chinese and Japanese fonts
Right now, font-family declarations look like this:

```css
font-family: "Noto Sans", "Noto Sans CJK JP", "Yu Gothic", "Hiragino Sans", TakaoPGothic, "PingFang SC", "Microsoft YaHei", 微软雅黑, sans-serif
```

Here we're using fallbacks to render the right font in the right language under the assumption that all glyphs for that language will be rendered by the first available font.

The problem is that Japanese fonts use a subset of Chinese glyphs. As a result, for Windows users who have both `Yu Gothic` and `Microsoft YaHei` installed, in a single block of text, both fonts end up getting rendered:

Example from a [Window Machine Running Microsoft Edge](https://whatismybrowser.com/w/7LXAJKW).

54f29542-38d4-4162-bb1d-d119bbc0085e

344a59a9-3824-4c89-b84c-6f2b2d2295f7

It's possible that these two font-face declarations can't live side by side and that we'll have to create a separate font-face file for Chinese and Japanese, the way we do for Arabic.

justintemps:
Fixes #986 

The problem was that in the `font-family` declarations, we were asking for Japanese system fonts before Chinese system fonts. Because Japanese system fonts include a subset of Chinese characters, those glyphs were getting rendered instead of the Chinese glyphs for the same character. Asking for the Chinese system fonts first systems to fix this without breaking Japanese font rendering.

I've only tested this on a Windows machine with Chrome (which is where we were having the problem, so it needs to be confirmed by Japanese and Chinese users on other devices). cc @inesdgomes 

## How to test

1. [Go to this website](https://pinkylam.me/generator/chinese-lorem-ipsum)
2. Copy all of the text that appears
3. [Go to the Twig deploy preview of this PR](https://deploy-preview-1089--ilo-ui-twig.netlify.app/?path=/docs/components-content-rich-text--rich-text)
4. Select `Canvas` in the top left 
5. Paste the Chinese text the content window of the controls menu at the bottom

Screenshot 2024-06-28 at 14 39 23

### Chinese RichText Before

RichText-Chinese-Before

### Chinese RichText After

RichText-Chinese-After

## Chinese Card Before

Cards-Chinese-Before

## Chinese Card After

Cards-Chinese-After

### Japanese RichText Before

RichText-Japanese-Before

### Japanese RichText After

RichText-Japanese-After

### Japanese Card Before

Card-Japanese-Before

### Japanese Card After 

Card-Japanese-After

### DIFF
--- a/.changeset/old-chairs-return.md
+++ b/.changeset/old-chairs-return.md
@@ -1,0 +1,6 @@
+---
+"@ilo-org/styles": patch
+"@ilo-org/themes": patch
+---
+
+Fixes a bug where Japanese characters were getting rendered instead of Chinese characters for some users.
--- a/packages/styles/scss/theme/_foundation.scss
+++ b/packages/styles/scss/theme/_foundation.scss
@@ -6,11 +6,10 @@
     * Typography
     */
   --ilo-scale: 1; // default scale for the design system
-  --ilo-font-family: Overpass, Noto Sans, Noto Sans CJK JP, Yu Gothic,
-    Hiragino Sans, TakaoPGothic, PingFang SC, Microsoft YaHei, 微软雅黑,
-    sans-serif;
-  --ilo-font-copy: Noto Sans, Noto Sans CJK JP, Yu Gothic, Hiragino Sans,
-    TakaoPGothic, PingFang SC, Microsoft YaHei, 微软雅黑, sans-serif;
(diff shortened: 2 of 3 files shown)


justintemps:
This won't work, it causes the same problem in Japanese that we had in Chinese.

Issue #1045, opened 2024-06-07
eslint-config: Upgrade to ESLint v9.0.0 


justintemps:
More details here:
https://eslint.org/blog/2024/04/eslint-v9.0.0-released/

This is blocked by #1038 as Eslint 9.0.0 has dropped support for everything < Node 20

PR #1093, opened 2024-07-02
Maestro 

GGKapanadze:
# Description 

The original name for this package was planned as `@ilo-org/composer`, but to avoid any confusion with `php`-s composer it was renamed as maestro  🧑‍🎨

This implementation provides a scoped package  that allows `storybook` to support  
- twig rendering 
- js bundling (cross-component)
- YAML parsing 
- UI Patterns schema (`wingsuit` compatible) including **variants** 
- Code snippet generation for preview 

##  Topology 

The `@ilo-org/maestro` package has three main concepts that are also expressed in package exports 

image

- `@ilo-org/maestro/plugin` - exports the vite plugin that allows all transformation 
- `@ilo-org/maestro/sb` - is a collection of storybook helpers, like decorators 
- `@ilo-org/maestro` - the main entry point that exports the core `Maestro` object 

### Plugin 
As mentioned above, the `@ilo-org/maestro/plugin` exports `UIPattern` vite plugin that accepts the following arguments

```ts
type Options = {
    namespaces: Record<string, string>;  // Twig namespaces
    globals?: Record<string, unknown>;   //  Globally Injected variables
};
```

The usability is dead simple 

````js
//vite.config.js
import { defineConfig } from "vite";
import { join } from "node:path";

import { UIPatterns } from "@ilo-org/maestro/plugin";

export default defineConfig({
  plugins: UIPatterns({
    namespaces: {
      components: join(__dirname, "..", "components"),
    },
    globals: {
      prefix: "ilo",
    },
  }),
});
````

### SB
`@ilo-org/maestro/sb` has utility helpers right now it has two

- `head` - the function that injects Drupal globally for each story 
- `BehaviorDecorator` - the storybook decorator that allows all Drupal behavior support 

> Example

```js
import { BehaviorDecorator } from "@ilo-org/maestro/sb";

/** @type { import('@storybook/html').Preview } */
const preview = {
  decorators: [BehaviorDecorator],
  parameters: {
    ...
  },
};

``` 
### Main

`@ilo-org/maestro` exports only one object named `Maestro.` Maestro has `create` method and that accepts the following:

- component
- pattern definition 

And in return you get the object 
```js
{
  meta: {...}
  stories: [...],
}
```

The stories key will always have 1 element, which is the `Default` story, and if the pattern definition has any variants, you will get all other stories in an array 

> Story Example

```js
import Accordion from "../components/accordion/accordion.twig";
import AccordionPatterns from "../components/accordion/accordion.pattern.yml";
import { Maestro } from "@ilo-org/maestro";

const story = Maestro.create(Accordion, AccordionPatterns);

const Meta = {
  title: "Components/Content/Accordion",
  tags: ["autodocs"],
  ...story.meta,
};

const [Default, Scrollable, Focus] = story.stories;

export default Meta;

export { Default, Scrollable, Focus };
```

## Variants

There are a few rules for variants: 
- Variants will inherit any setting from the default definition 
- Variants named `default` will be ignored because `Default` is generated by the pattern definition itself 

## Components
- Link
- Linklist
- Accordion

The following components were migrated during the development 

## Demo

https://github.com/international-labour-organization/designsystem/assets/36326203/5d1f221b-15a5-4883-bed4-95f4c9f22089


packages/maestro/.eslintrc
```diff
@@ -0,0 +1,46 @@
+{
```
bashlk, line 1: Isn't there a shared eslint config for all the packages in the monorepo?

packages/maestro/src/generator/definitions.ts
```diff
@@ -0,0 +1,40 @@
+// UIP = UI Patterns
```
bashlk, line 1: It's better not to abbreviate things for the sake of clarity. In this case, UIPatternFieldTypes is not such a long name.

packages/maestro/src/generator/schema.ts
bashlk, line 1: This checking of the definition structure is a really good addition. But this only runs when the Maestro is invoked to generate the stories right? Does it error out if the definition does not match?

packages/maestro/src/generator/story.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  if (!pattern.variants) {
+    return [defaultStory];
+  }
+
+  const stories: StoryObj[] = [defaultStory];
+
+  for (const [key, variant] of Object.entries(pattern.variants)) {
+    if (key === "default") continue;
+    const merged = Object.assign({}, variant.fields, variant.settings);
```
bashlk, line 35: You could use the spread syntax instead of Object.assign.

packages/maestro/src/plugin/twig.js
```diff
@@ -0,0 +1,276 @@
...
+ * TODO: reimplement with
+ *   - more typesafe approach
+ *   - remove react support
+ *   - decompose
+ */
+
+import { resolve, dirname } from "node:path";
+import { existsSync } from "node:fs";
+
+import Twig from "twig";
```
bashlk, line 13: I am concerned that this implementation of twig will behave differently from the twig implementation used in the PHP ecosystem. Already it notes that a few things that behave differently - https://github.com/twigjs/twig.js/wiki/Implementation-Notes#differences--implementation-notes

We had this risk before with wingsuit as well I think since it also leveraged a node implementation of twig.

packages/maestro/src/plugin/twig.js
```diff
@@ -0,0 +1,276 @@
+/**
+ * This is a transferred version of `larowlan/vite-plugin-twig-drupal
+ *
+ * TODO: reimplement with
```
bashlk, line 4: Transfer these improvements to a low priority issue so that we can eventually get to it. Add as much context as you can since it is fresh in your mind right now and it won't be in the future.

packages/maestro/.eslintrc
GGKapanadze: Yep, there is, but the `typesript` 

- Doesn't respect prettier config
- Has a bit of outdated configuration (wrong project resolution )
- Isn't strict enough 

We talked with @justintemps about config refresh because it's 2 years old. If we like the new ruleset, we can make it a new default.

packages/maestro/src/generator/schema.ts
GGKapanadze: We are doing the runtime check, so yes, an incorrect shape will have a well-documented error 

image

packages/maestro/src/generator/story.ts
GGKapanadze: I prefer to use `Object.assign` when I'm just merging the objects. Spread is a better tool for modifying or adding a property, but not for merging; `Object.assign` is more declarative and scalable

packages/maestro/src/generator/types.ts
```diff
@@ -0,0 +1,59 @@
+import { UIPatternValue } from "./schema";
+import { UIPFieldToStorybookType, UIPFieldTypes } from "./definitions";
+
+import { MaestroStory } from ".";
+
+type ArgTypes = MaestroStory["meta"]["argTypes"];
+
+function types(pattern: UIPatternValue): ArgTypes {
```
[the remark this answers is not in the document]
GGKapanadze: If you check the `create.ts`, it uses the `types` utility to generate  `CSF` [argTypes](https://storybook.js.org/docs/api/arg-types),

packages/maestro/src/plugin/twig.js
GGKapanadze: [Context](https://github.com/international-labour-organization/designsystem/issues/1046#issuecomment-2169653852)

I get the point, but it's the best we have in the `node` ecosystem. What's your alternative?

packages/maestro/src/plugin/twig.js
```diff
@@ -0,0 +1,276 @@
+/**
+ * This is a transferred version of `larowlan/vite-plugin-twig-drupal
```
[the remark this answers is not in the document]
GGKapanadze: We have [forked](https://github.com/international-labour-organization/twig-processor) it but maintaining separate repos will be an additional pain. Since it's small enough, we decided to include it as a part of `maestro`

packages/twig-new/.storybook/main.js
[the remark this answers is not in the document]
GGKapanadze: It's a reasonable idea. We can discuss it during the meeting

Shashika6:
Had a look a this one :') Will be ready to help with the migration once this gets merged

packages/maestro/src/generator/schema.ts
bashlk: Does an error in the definition cause the build process to fail?

packages/maestro/src/plugin/twig.js
bashlk: That is why I suggested switching to using the UI patterns module in Drupal as our dev playground. With the docker image, running it is trivial and already the docker image is being used to run the twig tests so it is not a new addition.

But as we previously discussed, I don't know whether hot reloading would work and that environment is not as interactive as storybook. So this is a risk we have to live with I guess.

packages/maestro/src/plugin/twig.js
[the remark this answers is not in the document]
bashlk: @justintemps I thought it would be useful to have a copy of the original source somewhere since we are going to modify this copy of it and because we can't trust that the original repo would stay up.

But I guess this plugin is our code now so the original source is of little help.

packages/maestro/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./lib/index.js"
+    },
+    "./plugin": {
+      "import": "./lib/vite.js",
+      "require": "./lib/vite.js"
+    },
+    "./sb": {
```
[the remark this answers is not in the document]
justintemps: Agreed, @GGKapanadze  let's make this change.

packages/maestro/src/plugin/twig.js
justintemps: Yup, @GGKapanadze could you do this?

packages/maestro/.eslintrc
justintemps: @GGKapanadze and @bashlk I'm ok with this, we can leave it here for now and then use it as a starting point when we tackle #1045

packages/maestro/src/generator/definitions.ts
justintemps: I agree, @GGKapanadze please make this change

justintemps:
Hey @GGKapanadze I went through the reviews, closed the ones we aligned on and left the ones that still require your action. Could you follow up on this and then merge when you're ready?

packages/maestro/src/generator/schema.ts
GGKapanadze: Again we, are doing runtime validation, not build time

bashlk: Sorry, it wasn't very clear from your previous reply.

I think mismatched patterns should cause the build to fail. or that we at least run a check for the pattern definitions on the CI. Could you create a separate task for this?

justintemps:
Hey @GGKapanadze, I tried to rebase `maestro` on top of `develop` but the lock files are out of sync and when I tried to fix them, kept getting an error to do with the patch. Weren't we going to replace that with a local version of the lib?

Anyway, for now I've successfully merged `develop` into `maestro` which muddies the git history somewhat, but I think it's ok. All good for you?

justintemps (approved):
@GGKapanadze we still need to replace the patch with a local version of the vite lib, but otherwise all good.

justintemps:
@bashlk @GGKapanadze @ademarco @Shashika6 here we go!

PR #1099, opened 2024-07-05
Use Foundation Font Families

### LINKED ISSUE #986: Fonts: Chinese text mixes Chinese and Japanese fonts

justintemps:
## Use Foundation tokens for font family declarations

Before we were using a mixture of css variables and a css mixin to set fonts in components and typography. This replaces all of that with a single reference to one of two variables set in the foundations file:

```scss
var(--ilo-fonts-display)
var(--ilo-fonts-copy)
```

## Use variables to set Chinese and Japanese System Fonts

In doing that it all fixes #986 

As explained in the issue, before we trying to use fallbacks to load Chinese and Japanese system fonts under the assumption that glyphs were each language were exclusive. That led us to a font-family declaration like this.

```
font-family: "Noto Sans", "Noto Sans CJK JP", "Yu Gothic", "Hiragino Sans", TakaoPGothic, "PingFang SC", "Microsoft YaHei", 微软雅黑, sans-serif;
```

Unfortunately, there is some overlap in glyphs between the two languages and so this results in characters getting rendered in different system fonts. 

With this PR, we're adding four new foundation variables

```scss
// Chinese system fonts
--ilo-fonts-display-zh: PingFang SC, Microsoft YaHei, 微软雅黑, sans-serif;
--ilo-fonts-copy-zh: PingFang SC, Microsoft YaHei, 微软雅黑, sans-serif;

// Japanese system fonts
--ilo-fonts-display-jp: Noto Sans CJK JP, Yu Gothic, Hiragino Sans, TakaoPGothic, sans-serif;
--ilo-fonts-copy-jp: Noto Sans CJK JP, Yu Gothic, Hiragino Sans, TakaoPGothic, sans-serif;
```

Then, we check the lang attribute set on the html element and if it's set to either one of those two languages, we reset the `--ilo-fonts-display` and `--ilo-fonts-copy` using the Chinese and Japanese font variables.

```scss
// Set Chinese system fonts
html[lang="zh"] {
  --ilo-fonts-display: var(--ilo-fonts-display-zh);
  --ilo-fonts-copy: var(--ilo-fonts-copy-zh);
}

// Set Japanese system fonts
html[lang="jp"] {
  --ilo-fonts-display: var(--ilo-fonts-display-jp);
  --ilo-fonts-copy: var(--ilo-fonts-copy-jp);
}
```

## Notes

We'll need to document this as Chinese and Japanese are the only languages whose fonts will be set according to the html language attribute.


GGKapanadze (approved):

packages/react/.storybook/styles.scss
```diff
@@ -5,8 +5,9 @@ h1,
 h2,
 h3,
 h4,
-h5 {
-  font-family: "Overpass", sans-serif !important;
+h5,
+h6 {
```
bashlk, line 9: Are these declarations even necessary since they are being defined in monorepo?

packages/styles/scss/_typography.scss
```diff
@@ -10,35 +10,32 @@
...
 }
 
-/// Include the `font-family` definition for the given name in your selector
-/// @param {String} $name
-/// @access public
-/// @group @carbon/type
-@mixin font-family($name) {
-  font-family: font-family($name);
+// Set Japanese system fonts
+html[lang="jp"] {
```
bashlk, line 36: Was the design system tested with all the languages that it should support? I have a feeling that there can be similar issues in other languages

packages/styles/scss/_mixins.scss
```diff
@@ -427,7 +427,7 @@
 
   li {
     @include font-styles("body-small");
-    font-family: $fonts-copy;
+    font-family: var(--ilo-fonts-copy);
```
bashlk, line 430: fonts-copy is for the text and fonts-display is for the headings?

packages/react/.storybook/styles.scss
justintemps: Unfortunately yea since Storybook has its own font styles on pages that you have to override.

packages/styles/scss/_typography.scss
justintemps: @bashlk it was tested with Latin languages, Cyrillic languages and Arabic. Chinese and Japanese weren't tested adequately. What makes them different is that you don't ship Chinese and Japanese fonts, they have tens of thousands of glyphs so webfonts aren't really a thing for those languages. You're stuck using system fonts in those cases, unlike languages like Arabic where you have a discreet character set that you can support with webfonts.

bashlk (approved):

justintemps:
It looks like the website is using `$fonts-display` so we may have to put that in to prevent the breaking change. Will need to check `$fonts-copy` as well.

PR #1124, opened 2024-07-16
feature(twig): implemented build scripts using vite

### LINKED ISSUE #1115: Twig Migration: build and export lib
Once we'll have migrated all of the components, we need to recreate the step that builds and exports the library for distribution.

bashlk:
Fixed #1115 

This PR implements the build script for the new twig package using vite. It is now around 6x faster than the previous implementation.

Before
```
pnpm build:lib  30.63s user 2.32s system 161% cpu 20.359 total
```

After
```
pnpm build:lib  5.49s user 0.91s system 136% cpu 4.695 
```

Compared to the previous twig package, there are a few notable changes
- The JS is output as ES modules instead of self contained UMD modules - this means that some of the JS files import a shared chunk file with common code. As long as all the JS files are hosted with the same directory structure as exported in the package, this should work fine in the browser.
- The build output follows the project structure - the current implementation deals with sub components in an inconsistent way. For example for the `tag` sub component that is under `tags` in the project, the twig file will be copied to `tags` and then the pattern and css files are copied to `tag`. In the new implementation, the output files will be follow the structure of the project so in this case `tag` will be placed under `tags`.
- Component CSS files are no longer outputted - The CSS file names don't match the names of the components in the twig package so they can't correctly be colocated in the build output. Since both the base theme and ilo.org site use the single combined style file, the twig package only outputs that now.


bashlk:
@GGKapanadze @justintemps I have an uneasy feeling switching to ES modules. It has great browser support but it is still not 100%. On one hand, shared code is not duplicated. But on the other, the output modules are not fully minified and as far as I can see, vite / rollup can't be made to output minified ES modules.

If we choose not to use ES modules, we have to switch back to using pure rollup. Vite doesn't support outputting several UMD / IIFE modules so we have to go back to feeding several configs to rollup.

bashlk:
After giving it some thought, I think it is best if we bundle each behavior file separately as we do with the current twig package. That will yield completely standalone JS files for each component that are the most minified and the most compatible across all browsers.

Sadly, this means that we will have to continue using rollup but I think I can improve the build scripts used in the old twig package. Will make the adjustments as soon as I can.

bashlk:
However, on the other hand, the JS files from the twig package will be processed by Drupal before they are sent to the user (https://www.drupal.org/node/3301716) and it seems Drupal supports ES modules for its JS files. It apparently also minifies the JS files 🤔

So I will test how ES modules work in Drupal. If it works, we can leave this PR as is.

ademarco:
@bashlk I've tested it and we can use ES modules in Drupal.

bashlk:
After exploring this a bit further with @ademarco we think that ES modules is not the way to go.

When a behavior JS file is marked as a module on Drupal, Drupal skips the aggregating the JS and instead requires the client to download each behavior file separately which is not performant. This combined with less browser support and the larger size of each behavior file when output from Vite as a ES module, leads it to be a poor choice.

So we proceed with iife. cc @GGKapanadze 

(umd is a combination of several formats which we don't need - https://rollupjs.org/configuration-options/#output-format - so iife is the way to go)

packages/twig-new/vite.config.ts
```diff
@@ -0,0 +1,135 @@
...
+import terser from "@rollup/plugin-terser";
+import swc from "@rollup/plugin-swc";
+import { globSync } from "glob";
+import chalk from "chalk";
+
+import theme from "@ilo-org/themes/tokens/theme/base.json" assert { type: "json" };
+
+const behaviors = globSync("src/components/**/*.behavior.js");
+
+export default defineConfig({
```
bashlk, line 15: If we are only using Vite for this plugin, I would rather get rid of it and use only rollup for the build process to simplify things

packages/twig-new/vite.config.ts
```diff
@@ -0,0 +1,135 @@
...
+  const startTime = Date.now();
+  const inputOptions: RollupOptions = {
+    input: options.input,
+    plugins: [
+      nodeResolve(),
+      commonjs(),
+      swc({
+        swc: {
+          env: {
+            targets: "last 2 versions, not dead, > 1% in US, ie 11",
```
bashlk, line 111: I think we can drop IE11, what do you think @justintemps?

packages/twig-new/vite.config.ts
bashlk: That way you can expose an array of config to rollup, one each for the behavior files and one dummy config that contains the copy plugin.

packages/twig-new/vite.config.ts
```diff
@@ -0,0 +1,135 @@
...
+        const grey = chalk.hex("#A5ABB6");
+        console.log(
+          `${chalk.green("✄")} ${chalk.blue("vite:")} ${grey("copied: ")}`
+        );
+        console.log(chalk.green("  - .twig files"));
+        console.log(chalk.green("  - .component.yml files"));
+        console.log(chalk.green("  - global styles\n"));
+
+        for (const behavior of behaviors) {
+          build(behavior);
```
bashlk, line 70: It's a bit confusing that the actual build happens in the buildEnd step of Vite.

packages/twig-new/vite.config.ts
GGKapanadze: Our build pipeline looks something like this.

image


 Since we have the `first phase` that should be triggered before bundling the JS, I think Vite can play a good role there. We can utilize the rich `plugin` system in the future and organize any pipeline operators into plugins. If we ditch `vite` and move to the rollup completely, then we should implement the `glue` step

packages/twig-new/vite.config.ts
GGKapanadze: I can agree if someone wants to support `IE 11` they can re-bundle during their build step

packages/twig-new/vite.config.ts
GGKapanadze: Why? We have the Vite  plugin that hooks into lifecycle of the build step

packages/twig-new/vite.config.ts
```diff
@@ -0,0 +1,135 @@
...
+          /(.*)\/(.*)\/(.*).behavior.js$/
+        )?.[2];
+
+        return `components/${componentFolder}/${chunkInfo.name}.js`;
+      },
+    },
+  };
+}
+
+async function build(id: string) {
```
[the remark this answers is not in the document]
GGKapanadze: Yep, but we are using rollup js API, not the `cli`. It's the same thing but done by the code,

packages/twig-new/vite.config.ts
justintemps: @bashlk and @GGKapanadze Yea we can definitely drop support for IE11

bashlk:
@justintemps @GGKapanadze I reworked the build scripts to use only rollup, output IIFE modules and to output all the component CSS files to a separate folder called styles. Let me know if it looks alright to you.

packages/twig-new/rollup.config.js
```diff
@@ -0,0 +1,79 @@
+import { globSync } from "glob";
+import { nodeResolve } from "@rollup/plugin-node-resolve";
+import commonjs from "@rollup/plugin-commonjs";
+import swc from "@rollup/plugin-swc";
+import copy from "rollup-plugin-copy";
+
+import theme from "@ilo-org/themes/tokens/theme/base.json" assert { type: "json" };
```
justintemps, line 7: Hey @bashlk, just a note here, if we're eventually going to deprecate this package, where else should this value be set?

bashlk: A quick search of the repo shows that this value is also used to substitute the prefix in the react package and also in the storybook of this package.

Issue #1111, opened 2024-07-12
Maestro: vite-twig plugin should process dynamic imports


justintemps:
This is blocking #1106. More information in #1110

Maestro's vite-twig plugin can't handle dynamic imports like this:

```twig
{% block modal_content %}
    {% include '@components/' ~ component ~ '/' ~ component ~ '.twig' with modaldata %}
{% endblock %}
```

We're using this pattern in at least two components:

- [ ] tabs
- [ ] modal

Until Maestro can handle this, we won't be able to migrate these two components.

PR #1125, opened 2024-07-18
Twig Dynamic Imports

GGKapanadze:
## Dynamic Imports

As mentioned in #1111, Mastro was unable to process the component that had imports like this 
```twig
{% include '@components/' ~ component ~ '/' ~ component ~ '.twig' with modaldata %}
```

The problem was that `vite-plugin-twig-drupal,` which is used under Maestro's hood, does not implement parsing the dynamic imports.

This PR allows Maestro to support imports like this.

## API

During the Maestro's integration, we have a new option named `dynamics`. This is a string array that should contain all component names that are using a dynamic import. 

```ts
export default defineConfig({
  plugins: UIPatterns({
    namespaces: {
      components: join(__dirname, "..", "src/components"),
    },
    globals: {
      prefix: "ilo",
    },
    dynamics: ["modal", "tabs"],
  }),
});
```

This helps the maestro prevent `cross-dynamic` imports. For example, if we have to support dynamic imports, we need to prepopulate the components that may be imported during development but exclude those that also have dynamic imports.

### Demo

https://github.com/user-attachments/assets/75a1e570-1e90-4136-bb29-b86ae98c5000

### DIFF
--- a/packages/maestro/src/plugin/twig.js
+++ b/packages/maestro/src/plugin/twig.js
@@ -8,7 +8,7 @@
  */
 
 import { resolve, dirname } from "node:path";
-import { existsSync } from "node:fs";
+import { existsSync, readdirSync } from "node:fs";
 
 import Twig from "twig";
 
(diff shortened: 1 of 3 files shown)


bashlk:
@GGKapanadze Can you elaborate on what you mean by cross dynamic imports? I don't understand why the imports have to be restricted a predefined subset of components.

packages/maestro/src/plugin/twig.js
```diff
@@ -151,6 +175,15 @@ const plugin = (options = {}) => {
           frameworkInclude = `import React from 'react'`;
           frameworkTransform = `const frameworkTransform = (html) => React.createElement('div', {dangerouslySetInnerHTML: {'__html': html}});;`;
         }
+
+        let isDynamic = false;
+
+        if (options.dynamics) {
```
bashlk, line 181: Would this work if a component that is specified in dynamics is used standalone?

GGKapanadze:
@bashlk  @justintemps 

Since we are in the storybook realm, to support dynamic imports, we need to preload other components beforehand, and this should be done because a user can hot-swap which component is rendered as a child. To detect if the current component is dynamic can be done by parsing the `AST`,  but the question is, why do we have a `dynamic` property?

We need to know all dynamic components, ahead of time, not to preload those components, since we don't have the implementation where dynamic components import other dynamic components

bashlk:
So the dynamic components are not preloaded and all the other components are, so that when a dynamic component starts, it can reference any of the other preloaded components?

And this setup breaks if we have dynamic components that import other components?

PR #1133, opened 2024-07-23
build(twig): restored package.json fields

### LINKED ISSUE #1116: Twig Migration: Update package.json
We need to make sure that the package.json of the new Twig project is consistent with the one before. It should have more or less the same fields, scripts, version number, etc.

bashlk:
Fixes #1116 

This PR restores all the fields in package.json to be similar to the old twig package. Since the lint commands did not run without the config, this PR also restores and adapts the ESLint config and fixes the lint errors that it found.

⚠️ This should probably be merged just before the old twig package is removed. This PR assumes that several scripts will be added in other PRs.

### DIFF
--- a/packages/twig-new/package.json
+++ b/packages/twig-new/package.json
@@ -46,8 +46,8 @@
   ],
   "scripts": {
     "build:lib": "rollup -c",
+    "build:docs": "storybook build -o ./storybook-static",
     "storybook": "storybook dev -p 6006",
-    "build-storybook": "storybook build",
     "start:theme": "docker compose up -d",
     "cy:open": "cypress open",
     "test": "cypress run",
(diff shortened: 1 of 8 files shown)


packages/twig-new/package.json
```diff
@@ -1,14 +1,57 @@
...
+    {
+      "name": "Giorgi Kapanadze",
+      "email": "doublegkapanadze@gmail.com",
+      "url": "https://github.com/ggkapanadze"
+    }
+  ],
   "scripts": {
     "storybook": "storybook dev -p 6006",
-    "build-storybook": "storybook build"
+    "build-storybook": "storybook build",
```
GGKapanadze, line 46: I think we need to change the output directory to support the current deployment workflow, the old one was outputted in `/storybook-static` 

Google Chrome 2024-07-25 13 40 34
https://storybook.js.org/docs/api/cli-options

PR #1138, opened 2024-07-26
Node 20 🎉

GGKapanadze:
# Description

After the maestro's introduction, we can upgrade from `node:16` to `node:20`. This allows us to remove extra `prep.` scripts during the installation. PR also updates the version of node for the github workflows

| Command | Tested |
|--------|--------|
| `pnpm build:libs`  | ✅  |
| `pnpm twig:dev:docs` | ✅ |
| `pnpm twig:build:docs`  | ✅ |
| `pnpm react:dev:docs` | ✅ | 
| `pnpm react:build:docs` | ✅ | 

> `pnpm build:libs` covers all internals


.nvmrc
```diff
@@ -1 +1 @@
-v20.11.1
+v20
```
bashlk, line 1: Just to be safe, might be better to set this to the same exact version as the workflows

package.json
```diff
@@ -27,7 +27,8 @@
     }
   ],
   "engines": {
-    "pnpm": "8.6.0"
+    "pnpm": "8.6.0",
+    "node": ">=20"
```
bashlk, line 31: Same note here

.nvmrc
GGKapanadze: Do you want to freeze `patch` versions, too?

bashlk: I think so, so as to keep it consistent. Then a single search for the version number will also reveal all the places in which it has to be updated.

bashlk (approved):

PR #1140, opened 2024-07-30
Twig 2.0

GGKapanadze:


.changeset/empty-dogs-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ilo-org/maestro": minor
+---
+
+story creation
```
bashlk, line 5: Should probably say initial version of the package right?

bashlk:
Fixed the build command in netlify settings for  ilo-ui-twig-develop. I didn't change it for the others because it would break the build until this branch is merged.

@justintemps let us know if we should update the build command for those sites as well or you can do it. (Just need to remove the non existing react:install and twig:install commands from the build command)

justintemps:
Thanks @bashlk great if you could do it.

bashlk:
@justintemps Corrected the build command for all the storybook sites. Build might now fail until the new changes are merged to main.

Issue #1141, opened 2024-07-31
Twig Docs: Use correct fonts in Storybook


justintemps:
It looks like the new Twig Storybook project is using incorrect fonts in several places. 

In general, we should apply the same rules to Storybook typography that we have in the design systems, using display fonts for headings and copy fonts for all other text. [See foundations](packages/styles/scss/theme/_foundation.scss). Only these fonts should be used, we shouldn't have text in Storybook rendered with different fonts.

I think the fonts were added in #1122 and #1123, but we need to override the font-family declarations for Storybook elements.

Here are some specific examples.

## Sidebar

In the old version of the Twig Storybook, these used Noto but I think they should use Overpass instead.

<img width="1709" alt="Screenshot 2024-07-31 at 10 14 24" src="https://github.com/user-attachments/assets/88d3324e-71ec-4d99-8ea2-ad36d9559173">

## Story content

<img width="1703" alt="Screenshot 2024-07-31 at 10 17 53" src="https://github.com/user-attachments/assets/8ecd0d75-c1a2-4284-af3a-cef0c4a0d9b3">

## Args tables content

<img width="1398" alt="Screenshot 2024-07-31 at 10 23 44" src="https://github.com/user-attachments/assets/0ac0ad3a-0f41-446f-9738-b6210c042065">

## Docs Pages

<img width="1422" alt="Screenshot 2024-07-31 at 10 25 42" src="https://github.com/user-attachments/assets/5e2cfe4f-1a3c-4816-9bff-eb7e5458f717">

PR #1149, opened 2024-08-06
docs(twig): refactor fonts to use overpass where needed

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/1141

Note :- 

Use Overpass in the following storybook elements.

- [ ] Sidebar
- [ ] Story Content
- [ ] Args Table
- [ ] Docs Pages

### DIFF
--- a/packages/twig/.storybook/manager-head.html
+++ b/packages/twig/.storybook/manager-head.html
@@ -1,8 +1,19 @@
 <!-- .storybook/manager-head.html -->
+<style>
+  @import url("/fonts/fonts-google.css");
+  @import url("https://fonts.googleapis.com/css2?family=Overpass&display=swap");
+  @import url("https://fonts.googleapis.com/css2?family=Fira Code&display=swap");
+</style>
 
 <!-- Canonical URL -->
-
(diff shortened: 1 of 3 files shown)


packages/twig/.storybook/styles.scss
justintemps, line 1: @Shashika6 we should use the foundation variables instead of SCSS vars from the soon to be legacy theme package.

packages/twig/.storybook/manager-head.html
```diff
@@ -1,8 +1,19 @@
 <!-- .storybook/manager-head.html -->
+<style>
+  @import url("/fonts/fonts-google.css");
```
justintemps, line 3: @Shashika6 Is it not possible to get these fonts from the fonts package?

justintemps (changes requested):

justintemps:
@GGKapanadze to check what Storybook recommends in terms of using locally hosted fonts.

Shashika6:
Did the changes to use css variables.

justintemps:
@GGKapanadze  and @Shashika6, I'm going to merge this and we can worry about how the fonts get imported into Storybook later.

PR #1163, opened 2024-08-11
React Storybook Upgrade

GGKapanadze:
## Description 

This pull request upgrades the storybook and its dependencies to the 8th version.

## Main refactor points

- `main` and `preview` configs were moved to object exports
- `ArgsTable` was refactored to `ArgTypes` 
- Typescript `declaration` omitting was disabled for development(storybook)
- `twig` and `react` share the same storybook addons 
-  Vite version was upgraded to `v5` (same as Twig)
-  Sidebar styles were synced with the twig 

## Scripts

| script | status |
| ------| ------ |
| build:lib| ✅ |
| build:docs | ✅ | 
| storybook | ✅ | 
| storybook:static | ✅ | 

## Demo

Demo of staticly built `build:docs` running with `storybook:static` 

https://github.com/user-attachments/assets/32dd85eb-c579-4b2e-89b8-f41541c1d78d


packages/react/.storybook/preview.tsx
```diff
@@ -1,81 +1,83 @@
...
-  },
-  viewMode: "docs",
-  layout: "padded",
-  docs: {
-    theme: theme,
-    source: {
-      excludeDecorators: true,
-    },
+export const globalTypes = {
+  rtlDirection: {
```
bashlk, line 9: Do we still need this with the new RTL plugin?

packages/react/.storybook/preview.tsx
```diff
@@ -1,81 +1,83 @@
...
+  decorators: [
+    (Story) => (
+      <GlobalProvider prefix={themeprefix.value}>
+        <Story />
+      </GlobalProvider>
+    ),
+    // Wraps components in story mode
+    (Story, ctx) => {
+      // Should this story be rendered in full width?
+      const fullWidthStories = ["footer", "localnav", "navigation", "hero"];
```
bashlk, line 25: @justintemps Do we really need different behavior implemented for different components?

packages/react/.storybook/theme.ts
```diff
@@ -1,12 +1,11 @@
-// .storybook/YourTheme.js
-
 import { create } from "@storybook/theming/create";
+//@ts-ignore
```
bashlk, line 2: Why ts-ignore? 

If it due to an issue in the brand assets package, lets create another issue to fix it and add a note here about why the ignore was added.

.changeset/small-actors-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ilo-org/react": patch
+---
+
+upgraded storybook to 8th version
```
bashlk, line 5: Specific version number here would be better.

packages/react/.storybook/preview.tsx
GGKapanadze: Good catch

packages/react/package.json
```diff
@@ -91,8 +91,7 @@
     "lint:fix": "eslint . --fix",
     "storybook": "storybook dev -p 6006",
     "storybook:static": "http-server ./storybook-static",
-    "copy:styles": "node copystyles.js",
-    "test": "NODE_OPTIONS=--experimental-vm-modules SKIP_PREFLIGHT_CHECK=true react-scripts test --watchAll=false --setupFilesAfterEnv ./src/setup.ts"
```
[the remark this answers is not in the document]
bashlk: NVM found commit that makes the change (chore(react): remove test script before tests migration). It would have been better if this change was done in the test migration PR.

packages/react/.storybook/preview.tsx
```diff
@@ -1,81 +1,53 @@
...
-      ctx.title.toLocaleLowerCase().includes(story)
-    );
-
-    if (ctx.viewMode === "story") {
-      if (isFullWidth) {
-        return <Story />;
-      }
-
-      return (
-        <div className="story-decorator-wrapper ">
```
bashlk, line 65: You can also remove the CSS for this decorator from preview-head.html

packages/react/package.json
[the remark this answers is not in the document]
GGKapanadze: It was done to avoid unexpected pipeline failures

PR #1173, opened 2024-08-21
Logo Grid: fix layout and size issues

### LINKED ISSUE #954: Logo grid: document and improve component
We developed a new component called 'logo grid' that is not properly documented. The component now exists in [storybook ](https://ilo-ui-twig.netlify.app/?path=/docs/components-brand-logo-grid--logo-grid) but not in Figma. The developers followed the style used on this page: https://www.figma.com/file/TnVsWy6VobbHzZdbHdx3Rb/ILO-Page-Templates?type=design&node-id=3796-41633&mode=design&t=GCP28TXoAdcuBckG-0 
Screenshot 2024-04-17 at 15 19 21

We need to create the component in Figma and document it correctly. Besides, we need to do the following changes:
- Increase the size of the logos' images. At the moment, the logos can be a maximum of 150px wide and 50px high. The component will render them as high and as wide as possible until they hit one of those two limits.
- Design for scenarios when CMS editors use only one or two logos

FYI @justintemps Assigning this to Beatriz but copying you so know it's coming up.

justintemps:
Fixes #954 

Logo Grid now uses the following logic:

## Default breakpoint (mobile)

Items take up full width

## Small breakpoint

Items take up half width

## Medium breakpoint on up

- If there are one or two items, each item takes 1/2 the available width
- If there are more than two items, each item takes up 1/3 of the available width


justintemps:
Hey @Shashika6, could you give this a look when you have a chance?

Shashika6:
Sure @justintemps

justintemps:
@Shashika6 I fixed the failing tests and added some. Now I'd like you to add two more tests please:

1. One test should make sure that when then are less than three items, the items take up half the available space
2. The second should make sure than when there are three items are more, the items take up a third of the available space

Make sure to account for the grid gap when checking the available space.

Shashika6:
Hey @justintemps I added a test which checks if the row has 3 items when there are 3 or more items. We cant use the grid gap cause there is.a padding right that gets added to the container so its not accurate. I followed another approach let me know what you think. 

I will get back into this further when I have configured the endpoint to pass in data in which we can do additional tests passing in 3 or less grid item data.

I will be starting on the endpoint refactoring and test expansion today.

packages/twig/cypress/e2e/media/logo-grid.cy.js
```diff
@@ -22,4 +24,39 @@ describe("logo grid", () => {
...
+        expect(itemWidth).to.equal(containerWidth / 2 - 16);
+      });
+    });
+  });
+
+  it("at large breakpoints ensure that there are 3 grid items in a row when there are 3 or more items.", () => {
+    cy.get(".ilo--logo-grid--logos").then(($container) => {
+      const containerWidth = $container.outerWidth();
+      cy.get(".ilo--logo-grid--item").should("have.length.gte", 3);
+      cy.get(".ilo--logo-grid--item").each(($item) => {
```
justintemps, line 55: Hey @Shashika6, this is one way to do it, but may be a more effective way to test:

1. That all of the items have the same width
2. That the width of all of the items is equal to 1/3 of the space available

Shashika6: Sure @justintemps I'll also do a check to see if all the items have the same width

justintemps: May be that was clear, so first you get the width of the all of the items and makes sure it's the same.  Then you do one single check to make sure that the value is equal to 1/3 of the space available.

Btw I'm pretty sure there's a better way to assert that a value is equal to three other than by asserting it has to be at least three but less than four...

Shashika6: Hey @justintemps this is the additional padding on the right I mentioned. 

test

Shashika6:
Hey @justintemps I refactored this to use grid hope its okay :), I also added proper tests that checks if the it width matches the item widths as expected as well.

justintemps:
Thanks @Shashika6 but I think you missed an important part of the spec here, which is that at the desktop breakpoint, when there are one or two items, they take up half the available width.

packages/twig/cypress/e2e/media/logo-grid.cy.js
```diff
@@ -22,4 +22,38 @@ describe("logo grid", () => {
         cy.get("img").should("have.attr", "src").and("not.be.empty");
       });
   });
+
+  it("at mobile breakpoint, shows items taking up full container width", () => {
```
justintemps, line 26: @Shashika6 no! Read the spec.

Shashika6:
Will fix this up @justintemps

packages/twig/cypress/e2e/media/logo-grid.cy.js
Shashika6: This one is for the mobile breakpoint @justintemps

Shashika6:
@justintemps I did the changes so that it satisfies 

Medium breakpoint

* If there are one or two items, each item takes 1/2 the available width

justintemps:
Thanks @Shashika6, now you need to add a test for what happens when there are one or two items at the desktop breakpoint.

justintemps:
Hey @Shashika6 I went ahead and added the test for when there are less than three items at the desktop breakpoint.

Issue #864, opened 2024-03-11
Multilink card: make image optional


inesdgomes:
We need the image in multilink cards to be optional. The designs need to be updated so the component accommodates this option. 
One important aspect is that we need to set a new logic for the card where the width has a narrower maximum size when there is no image, otherwise the text line length will be too long.

When the component is updated, we should also show it without the image in some of the templates in various breakpoints.

inesdgomes:
Hi @beatrizmartinmartins Assigning this task to you as it's one of the first things we'd like to address. We can talk about it on Monday.

inesdgomes:
@beatrizmartinmartins Could we aim for 22 April to have this ready?

beatrizmartinmartins:
@inesdgomes Done. I implemented the image as optional on the DS and created some templates with the multilink across different places and I think it works fine on the different breakpoints with what we talked about on our meeting. We only implement them on very few places. You can check those templates [here](https://www.figma.com/file/TnVsWy6VobbHzZdbHdx3Rb/ILO-Page-Templates?type=design&node-id=4884%3A95585&mode=design&t=cjmAwAxdGif8pRGI-1)

inesdgomes:
Thanks @beatrizmartinmartins Let's discuss this one tomorrow. A few questions that come to my mind:
- How do show/hide the image in the component library? I don't seem to be able to find that option in the Components' Figma file, though it is clear in the Templates file.
- Was the maximum width for the whole multilink card component changed? I know we talked about aligning it with the graphic promo's width, but I'm a bit concerned about the impact that will have on existing very long content in the website. I can explain in detail tomorrow.

beatrizmartinmartins:
Just created new variants for the Multilink Card. We have the With Image option that doesn't have a max-width, while the WithoutImage option has a max-width of 920px which is the size of the promo card. With these options and variants, we are able to decide what is the component we want and the constraints that come with it. What do you think @inesdgomes ? https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?m=auto&node-id=6389%3A11390&t=3cDu3JEAQ22cCyUi-1

inesdgomes:
Cool, makes sense to me, thanks! 
@justintemps We can move this to development. Here's what we need to do in the multilink card:
- Image becomes optional
- When there is **no image**, the component has a max-width of 920px (like the Promo card)
- When there is an image, the component does not have a max-width

justintemps:
[...]

PR #1177, opened 2024-08-23
fix(twig): make image optional in multilink card

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/864

### Notes

* Make image optional in multilink card
* Add max-width to wide and fluid when image is available. 

### Screenshot

Screenshot 2024-08-23 at 19 40 25

### DIFF
--- a/packages/styles/scss/components/_multilinkcard.scss
+++ b/packages/styles/scss/components/_multilinkcard.scss
@@ -125,6 +125,14 @@
                 display: none;
               }
             }
+
+            &#{$self}--no-image {
+              --max-width: #{px-to-rem(920px)};
+
+              #{$self}--content {
+                width: 100%;
+              }
+            }
           }
         }
       }
--- a/.changeset/blue-flies-report.md
+++ b/.changeset/blue-flies-report.md
@@ -1,0 +1,7 @@
+---
+"@ilo-org/react": patch
+"@ilo-org/styles": patch
+"@ilo-org/twig": patch
+---
+
+Make image optional in multilink card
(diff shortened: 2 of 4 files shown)


Shashika6:
@justintemps I noticed that you have used 

`  &__size__fluid {
    --max-width: 100% !important;
  }`
  
  in the card css which is overriding the size for the fluid to be 100% is this okay ?

justintemps:
> @justintemps I noticed that you have used
> 
> ` &__size__fluid { --max-width: 100% !important; }`
> 
> in the card css which is overriding the size for the fluid to be 100% is this okay ?

@Shashika6 yea, that's what gives the card fluid width to fill it's container

justintemps:
@inesdgomes could we get your eyes on this:

1. [Navigate here](https://deploy-preview-1177--ilo-ui-twig.netlify.app/?path=/story/components-cards-multi-link-card--default)
2. Click on the button to remove the image
Screenshot 2024-09-02 at 10 01 08
3. Check to make sure that it looks ok

packages/styles/scss/components/_multilinkcard.scss
```diff
@@ -125,6 +125,14 @@
                 display: none;
               }
             }
+
+            &#{$self}--no-image {
+              --max-width: #{px-to-rem(920px)};
```
justintemps, line 130: @Shashika6 what is this for? Why should there being an image or not affect the width of the card? Also, it doesn't seem to have any effect.

Shashika6: When there is no image, the component has a max-width of 920px. 

There was a requirement by design.

justintemps:
Hey @Shashika6 apologies, can you revert the change I asked for this morning? I discussed with @inesdgomes and we agreed that: max-width is different for the card when it doesn't have an image **but only when size is set to `wide`**. Otherwise, it's unaffected.

Shashika6:
Reverted the change again @justintemps

Issue #1176, opened 2024-08-23
Twig: Navigation searchInput should be selected by tag instead of class


justintemps:
In the Navigation component, right now we're selecting the searchInput like this:

```js
this.searchInput = this.searchBox.querySelector(`.${this.prefix}--input`);
```

The problem is that the search input appears inside a block...

```twig
{% block desktopSearch %}
    {% include '@components/form/form.twig' with {...}
{% endblock %}
```

Which means that the user can swap in their own input if they want to that might not have the className. That's what's happening on the website right now and it's causing a JavaScript error, because`this.searchInput` is undefined.

To fix this, we should simply do this:

```js
this.searchInput = this.searchBox.querySelector("input");
```

So no matter what input gets used, it will still get selected.

PR #1178, opened 2024-08-23
fix(twig) : refactor search input to be selected by tag

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/1176

### Notes

* Query search input by tag instead of class
* Modify unit test to check if input is focused after search button click
* Update base theme version to use 0.4.0

### DIFF
--- a/.changeset/silly-donkeys-play.md
+++ b/.changeset/silly-donkeys-play.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": patch
+---
+
+Refactor search input to be queried by tag instead of class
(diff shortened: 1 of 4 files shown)


justintemps:
Hey @Shashika6, just a reminder then when you write "Fixees :-" instead of "Fixes" it won't close the issue when I merge the ticket. The correct syntax is "Fixes #...."

Shashika6:
@justintemps can we get this one merged cause the CI is failing. I did the fix to update to the correct base theme version on this one.

Issue #1012, opened 2024-05-13
Update social media icon colors


inesdgomes:
As discussed, consider to use ILO blue (instead of grey) on social media icons in a white background.

https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?node-id=657-4040&t=JrKgQw0u3md7qsI6-0

beatrizmartinmartins:
Changed colours on [this component](https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?m=auto&node-id=657%3A4040&t=2KxVXrG6xwSm2oel-1) and [this component](https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?m=auto&node-id=7218%3A14778&t=2KxVXrG6xwSm2oel-1). Please check [this template](https://www.figma.com/design/TnVsWy6VobbHzZdbHdx3Rb/ILO-Page-Templates?m=auto&node-id=2499%3A48962&t=QLmjJtYWfagzLaEe-1) to see the change applied

inesdgomes:
@justintemps I can't see this change here: https://react.ui.ilo.org/?path=/docs/components-navigation-socialmedia--docs 

Please could you check?

PR #1179, opened 2024-08-25
fix(react-twig): refactor light social media icon color

Shashika6:
Fixes https://github.com/international-labour-organization/designsystem/issues/1012

### Notes

* Refactor social media icon color in light theme

### DIFF
--- a/packages/styles/scss/components/_socialmedia.scss
+++ b/packages/styles/scss/components/_socialmedia.scss
@@ -7,9 +7,10 @@
   $c: &;
   $default-theme: "light";
   $icon-size: $spacing-spacing-4;
+  $ilo-color-blue: "rgba(30, 45, 190, 1)";
 
   @mixin get-icon($icon, $theme: $default-theme) {
-    @include dataurlicon($icon, rgba(30, 45, 190, 1));
+    @include dataurlicon($icon, $ilo-color-blue);
 
     &:hover {
       @include dataurlicon(
--- a/packages/styles/scss/components/_socialmedia.scss
+++ b/packages/styles/scss/components/_socialmedia.scss
@@ -9,10 +9,7 @@
   $icon-size: $spacing-spacing-4;
 
   @mixin get-icon($icon, $theme: $default-theme) {
-    @include dataurlicon(
-      $icon,
-      map-get($color, "socialmedia", $theme, "icon", "color")
-    );
+    @include dataurlicon($icon, rgba(30, 45, 190, 1));
 
     &:hover {
       @include dataurlicon(
--- a/packages/styles/scss/components/_socialmedia.scss
+++ b/packages/styles/scss/components/_socialmedia.scss
@@ -7,10 +7,9 @@
   $c: &;
   $default-theme: "light";
   $icon-size: $spacing-spacing-4;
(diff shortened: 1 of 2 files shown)


packages/styles/scss/components/_socialmedia.scss
```diff
@@ -9,10 +9,7 @@
   $icon-size: $spacing-spacing-4;
 
   @mixin get-icon($icon, $theme: $default-theme) {
-    @include dataurlicon(
-      $icon,
-      map-get($color, "socialmedia", $theme, "icon", "color")
-    );
+    @include dataurlicon($icon, rgba(30, 45, 190, 1));
```
Shashika6, line 12: Hey team since we are moving away from map-get and using css variables, there a small problem in this one. Css variables cant be used in mixins. 

For the time being I used `rgba(30, 45, 190, 1)` as it doesn't support hex values as well. 

How should we go with this one @justintemps @bashlk @GGKapanadze

Shashika6:
@justintemps I tried passing the variable as a string but it doesn't seem to work. So for the time being I created a scss variable.

packages/styles/scss/components/_socialmedia.scss
```diff
@@ -7,9 +7,10 @@
   $c: &;
   $default-theme: "light";
   $icon-size: $spacing-spacing-4;
+  $ilo-color-blue: "rgba(30, 45, 190, 1)";
```
bashlk, line 10: I think what the guys meant is to assign the CSS variable to the SCSS variable so that it can be passed into the mixin.

Shashika6: You mean  $ilo-color-blue: var(--ilo-color-blue); right ? I tried this but it didnt seem to work.

justintemps:
@Shashika6 also I think you forgot to implement the dark theme version

justintemps:
@bashlk, @Shashika6 and @GGKapanadze, the problem here isn't that css vars can't be used in mixins, the problem is that they don't work inside a data url, which is what the icon function and mixins are doing. They give us something like this:

```css
background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 21'%3e%3cpath fill='var(--ilo-color-blue)' d='M18.901 0.153564H22.581L14.541 9.34356L24 21.8466H16.594L10.794 14.2626L4.156 21.8466H0.474L9.074 12.0166L0 0.154564H7.594L12.837 7.08657L18.901 0.153564ZM17.61 19.6446H19.649L6.486 2.24056H4.298L17.61 19.6446Z'/%3e%3c/svg%3e");
```

If you see, the variable is getting passed appropriately, but because the SVG isn't in the DOM per se, it doesn't work.

I think this reveals a small but important flaw in our new approach for handling tokens, in that css variables can't be used in some contexts. I'm not sure what the best way to handle this is.

GGKapanadze:
Yep, variables can't be directly used with fill, but if we still want to do background images for icons I can reimplement this mixin with 'mask-image'

Shashika6:
Marking this as a draft till https://github.com/international-labour-organization/designsystem/issues/1189 is completed.

Issue #985, opened 2024-04-23
React: Migrate to Vite and Vitest


bashlk:
Currently the React project uses a custom rollup configuration to build the assets, CRA to run the tests and Vite to build the components for storybook. This is a very strange setup since this causes the code to be built using multiple bundler configurations for each use case. It also adds a lot of dependencies to the project.

The React project should be migrated to use a single family of tools, the best candidate is probably Vite and Vitest. Vite has a [library mode](https://vitejs.dev/config/build-options#build-lib) for building React component libraries, this is probably what we need for our use case to replace the custom rollup config. I haven't come across a guide on their docs site for migrating an existing project to use Vite so we should probably adapt our scripts / configuration based on the [react-ts](https://vite.new/react-ts) template preset.

We already have Vite in the project since it is used by storybook but it is one major version behind. We should update it since Vitest requires the latest version. Vitest should be able to run the existing tests in the project out of the box but the tests should be manually checked. The basic guide for adding [vitest to a project](https://vitest.dev/guide/#adding-vitest-to-your-project) should suffice for this.

After the migration above, we should be able to remove CRA (`react-scripts`) and all the rollup plugins from the dev dependencies in the project (since they will be implicitly required by Vite). The configuration for jest in package.json could also be removed. 

I have also noticed several dev dependencies that are not being used (e.g. `nodemon`, `jest-environment-jsdom`, `react-element-to-jsx-string`). So the dev dependencies should be manually checked and the unused ones should be removed.

Shashika6:
Hey @justintemps did some initial work on this and @GGKapanadze will be picking this one up to optimise the current build/plugins

PR #1192, opened 2024-09-04
React build with Vite

GGKapanadze:
# Description 

> This PR is a part of #985 to migrate completely on `vite`, the second part will include `vitest` setup

From a high-level perspective, we are swapping out  bare `rollup` in favor of `vite`  but also

- New Dependency Strategy 
- We introduced a new external API
-  Optimised Bundle
- Upgraded `react` version 

## New Dependency Strategy 

Old Implementation, externalised every dependency during bundling phase, for example in order to use `video` component you have to install them separately, same goes for dependencies listed bellow

- classnames
- dom-helpers,
- nanoid
- react-transition-group,
- video.js
- videojs-youtube

The new approach only externalises `react` and `react-dom` so in order to start using `@ilo-org/react` you can just run

```sh
pnpm add react react-dom @ilo-org/react
```

## New External API

With the new bundle we have 2 main entry points:

- FIrst one covers the style part `@ilo-org/react/styles`, which should be used to import the styles
```jsx
...
import "@ilo-org/react/styles/global.css";
import "@ilo-org/react/styles/index.css";

createRoot(document.getElementById("root")!).render(
  <StrictMode>
    <App />
  </StrictMode>
);
```
- And the second one is just `@ilo-org/react` which is root for both `hooks` and `components` 

So in order to grab any `js` part from the library you can just import it from the root 

```ts
import { Loading, Button, useGlobalSettings  } from "@ilo-org/react";
```

typedefinitions are also exported to help with component interaction 

https://github.com/user-attachments/assets/622c0dd7-73bb-4108-b59f-1fc02defa7ac

## Optimised Bundle

New bundle is designed to be easily tree-shakable for many major bundlers.

- For example this is the size output for simple "Hello World" app with vite

image

- And this is the same output but with "Loading" component from "@ilo-org/react" 

image

- Loading + Button (that includes "Icon")

image

----

From development perspective `lib`dir looks like this 

image

We output both `esm` and `cjs` but by utilising the multiple input source for "rollup" we are outputin each individual files with optimised chunks that are stored inside `_store` 

 ## Upgraded `react` version 

And for the bonus, `@ilo-org/react` is now upgraded to `react:18` with fixed typeings


packages/react/package.json
```diff
@@ -146,37 +119,29 @@
...
     "tslib": "^2.3.1",
-    "vite": "^5.2.0"
+    "vite": "^5.2.0",
+    "vite-plugin-dts": "^3.9.1",
+    "vite-plugin-static-copy": "^1.0.6"
   },
   "peerDependencies": {
-    "react": ">16",
-    "react-dom": ">16"
+    "react": ">17",
```
bashlk, line 143: Since you upgraded to react 18, shouldn't these be changed to 18?

packages/react/vite.config.ts
```diff
@@ -0,0 +1,72 @@
...
+
+function names({ facadeModuleId }: PreRenderedChunk) {
+  if (facadeModuleId?.endsWith("src/index.ts")) {
+    return "index.js";
+  }
+
+  if (facadeModuleId?.includes("components")) {
+    const componentName = facadeModuleId.split("/").at(-2);
+
+    return `_store/${componentName}-entry-[hash].js`;
```
bashlk, line 18: Why is there _store in the path name?

packages/react/package.json
GGKapanadze: `>17`  will target everything above "17th" version 

image

packages/react/vite.config.ts
GGKapanadze: Just for the build structure, as mentioned in the PR description

packages/react/package.json
bashlk: Sure but >17 also includes all React 17 versions. Since we are using React 18 internally, I think we should set that as our peer dependency because the older versions could behave differently.

GGKapanadze: I will freeze it to the "18" then for readability

justintemps (changes requested):
Hey @GGKapanadze this all looks good to me, you just need to add a changeset. This looks like a minor release to me.

GGKapanadze:
Since there was no affected functionality, I marked it as a patch release

justintemps:
hey @GGKapanadze, we introduced a new external api and bumped react by two versions, I think that makes it a minor release :)

.changeset/young-lizards-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ilo-org/react": minor
+---
+
+upgraded types to support react 18
```
justintemps, line 5: @GGKapanadze we should mention how imports work now as well, no?

justintemps (approved):

PR #1216, opened 2024-09-18
Fix(Twig): backwards compatibility for social media component

justintemps:
This adds backward compatibility to the socialmedia component which in 1.1.1 removed the twitter icon and replaced it with X. With this, passing `twitter` to as an argument to the icon field will render an X icon.

Before merging it, we need to add some tests to the social media component. However, the tests that we have now are really testing the component, they're testing the arguments passed to the component from patterns, which is wrong.

I tried to update the tests using the ilo_base_theme pattern preview, but was unable. I've opened [#34](https://github.com/international-labour-organization/ilo_base_theme/issues/34) in that repo to track.

### DIFF
--- a/.changeset/ninety-cycles-cough.md
+++ b/.changeset/ninety-cycles-cough.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": patch
+---
+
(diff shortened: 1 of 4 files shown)


justintemps:
hey @Shashika6 please see the way I changed the tests here. I know that all of the tests so far are preliminary, but in too many cases we're testing the pattern more than the component. Now that we can pass arguments to the ilo_base_theme, we should use mock data to test edge cases rather then just the pattern data.

Issue #1198, opened 2024-09-09
Sync dependencies


justintemps:
After merging #1193, we get the following installation warnings

```shell
packages/icon-build-helpers
└─┬ react-dom 17.0.0
  └── ✕ unmet peer react@17.0.0: found 17.0.2

packages/icons-react
└─┬ react-dom 17.0.0
  └── ✕ unmet peer react@17.0.0: found 17.0.2

packages/maestro
└─┬ vite 5.3.1
  └── ✕ unmet peer @types/node@"^18.0.0 || >=20.0.0": found 17.0.45

packages/twig
└─┬ react-dom 18.3.1
  └── ✕ unmet peer react@^18.3.1: found 18.0.0

packages/utils
└─┬ ts-jest 29.2.5
  └── ✕ unmet peer jest@^29.0.0: found 27.5.1
```

PR #1221, opened 2024-09-20
Monorepo Maintance

GGKapanadze:
# Introduction

Everything started with #1045, but during the ESLint migration step, it was obvious that a mono-repo needed some maintenance, so this PR evolved to more general maintenance.

## Migration
As described in a #1045, we needed an upgrade to a new ESLint 9th version, which is a game changer for the whole js community you can check https://github.com/eslint/eslint/issues/18391, ESLint v8 [EOL](https://eslint.org/blog/2024/09/eslint-v8-eol-version-support/) is October 5, 2024 so it was perfect time for upgrade

A new `@ilo-org/eslint-config` is fully moved to flat configuration, but instead of one config we have 3

- `js` - this config is the base for everything, it introduces an opinionated `Javascript` config base  as long as opinions about formatting, with `eslint-plugin-prettier`, but it's not only for extending, we have a big `js` codebase that also needs lining and it's referenced in those packages
- `recommended` - this config was named `ts` during development, but since we are planning to move on `typescript` fully  I decided to name it `recommended`, it is based on the `js` config but adds `typescript-eslint` on top with opinionated rules
- `react` - since we have `react` design system, we also export config tailored for `react` that is based on `recommended` 
- `ignores`(internal) - instead of `.eslintignore` we have a separate config for ignoring files and directories 

image

### API
You can use any configuration you like by simply including it inside your eslint config

Just export it

```js
import configs from "@ilo-org/eslint-config";

export default configs.recommended;
```

or extend and add your own

```js
import configs from "@ilo-org/eslint-config";

export default [
  ...configs.recommended,
  {
    rules: {
      "no-console": "warn",
    },
  },
];
```

### Packages

This table displays which package uses which configuration:

| Package                | Configuration |
| ---------------------- | ------------- |
| `twig`                 | `js`  + internal `cypress`  config   |
| `icons`                | `js`          |
| `icons-react`          | `js`          |
| `icon-build-helpers`   | `js`          |
| `maestro`              | `recommended` |
| `utils`                | `recommended` |
| `brand-assets`         | `recommended` |
| `react`                | `react`       |

image

### Config Opinions

When I started working on configs, I had a more strict version of it in mind but I ended up building a bit "strict enough", to be more general, packages can add or remove any rules based on their contents, and also some legacy packages like `icon-build-helpers`  will be required

Here are a list of **opinions** inside the recommended config 

| Rule                                              | Status | Description                                                     |
| ------------------------------------------------- | ------ | --------------------------------------------------------------- |
| "prettier/prettier"                               | 🚨     | Is strong about formatting                                      |
| "no-console"                                      | 🚧     | Prevents the use of `console.log` but allows `warn` and `error` |
| "no-duplicate-imports"                            | 🚧     | Prevents duplicate imports                                      |
| "no-await-in-loop"                                | 🚧     | Prevents `await` inside loops                                   |
| "no-unused-vars"                                  | 🚧     | Prevents unused variables but allows `_` to be ignored          |
| "@typescript-eslint/class-literal-property-style" | 🚧     | Normalizes class literal properties                             |
| "@typescript-eslint/no-duplicate-enum-values"     | 🚧     |                                                                 |
| "@typescript-eslint/prefer-for-of"                | 🚧     | Enforces the use of `for-of`                                    |
| "@typescript-eslint/consistent-type-definitions"  | 🚧     | `interface` is preferred                                        |
| "@typescript-eslint/ban-ts-comments"              | ✅     |       Migrated from old config                                                          |
| "@typescript-eslint/ban-ts-ignores"               | ✅     |             Migrated from old config                                                      |

### Refactoring
After defining a new standard for the codebase, I started to adjust packages to it, I did a bunch of refactoring but I marked  13 files with a `@ESLINT-DEBT ` comment that should be refactored in the scope of the follow-up task

### TODOS
There are two plugins that I wanted to include inside of new configuration

- `eslint-plugin-import`
- `eslint-plugin-storybook` 

But due to lacking of eslint v9 support, those can also be added in the future

## Prettier & tsconfig

After eslint migration, it snowballed to a prettier and typescript config.

- Prettier is now upgraded to the latest version: `v3`
- TSConfig became more general, it no longer adds `dom` and `dom-iterable` by default, it is done on a package level 

Both of them are marked as a breaking change

## ESModules 

- Monorepo core
- eslint-config
- prettier-config

Now migrated to `"type": "module", and now they are fully utilizing ES Modules 

## Monorepo 

- `fix:pnpm:filter` is removed 
-  husky has its own install file 
-  unmet peer dependencies are fixed for all the packages #1198 
-  `format`, `format:fix` scripts are removed because eslint is converting the `prettier` rules with an error exit
- `brand-assets` package now has no type duplication 
-  config packages have updated docs 
-  unused dependencies were removed


prepare.js
```diff
@@ -1,10 +0,0 @@
-/**
- * This optionally loads husky except in CI environments
- * like Github Actions where the lint-staged pre-commit hook
- * seems to break. CI is always set to true in Github Actions
- * so there's no need to set it manually.
- */
-const isCi = process.env.CI !== undefined;
```
justintemps, line 7: Hey @GGKapanadze are you sure this is no longer an issue?

.eslintrc.js
justintemps, line 1: @GGKapanadze If you've got VS Code setup to take eslint and prettier defaults from the root of the project, it seems like this way it has nowhere else to get them? That was pretty useful for me, are we sure we want to get rid of that?

justintemps:
Hey @GGKapanadze this is awesome, I just had a few little questions and then I think this can be merged.

justintemps:
One more thing here, @GGKapanadze, running the lint and format commands on the monorepo edited about a dozen files. We should probably do that once and then commit before merging.

prepare.js
GGKapanadze: During my testing yes, I'll keep an eye on it

.eslintrc.js
GGKapanadze: we have `eslint.config.mjs` for the project root, which uses the default `js` config it will be triggered on any file that is have no package level eslint config and is not ignored 

https://github.com/user-attachments/assets/296e4882-9f18-454b-8771-6ab2e25afd70

GGKapanadze: try to restart eslint server for VSCode, to pick up a new references

`CMD` + `Shift` + `P` > ESLint: Restart Eslint Server

GGKapanadze:
@justintemps 

`lint` command outputs without any change, I'll commit the formatted styles

config/eslint-config/README.md
```diff
@@ -10,42 +10,55 @@ npm install -D @ilo-org/eslint-config
...
+`@ilo-org/eslint-config` provides a list of ESLint flat configurations:
 
-```json
-{
-  "extends": ["@ilo-org/eslint-config"]
-}
+| Config Name | Description                           | Internal Extends |
+| ----------- | ------------------------------------- | ---------------- |
+| js          | Base JavaScript config                | ⛔               |
+| recommended | Recommended configuration `js` + `ts` | `js`             |
```
bashlk, line 18: I still think that ts or typescript is a better name for this. recommended can mean anything and other opinions that are not related to typescript could also get added. By naming it typescript, we open the options of having more configs that build on top of typescript instead of the recommended getting adapted for every single use case.

From my understanding, you are proposing that project specific opinions should be placed as project level rules right?

config/eslint-config/README.md
```diff
@@ -10,42 +10,55 @@ npm install -D @ilo-org/eslint-config
...
+| react       | React Typescript config               | `ts` -> `js`     |
+
+You can use any configuration you like by simply including it inside your eslint config
+
+Just export it
```
bashlk, line 23: "Just export it" sounds a bit confusing since the config has to be imported first. I think it would be better if you say that it can be re-exported from the project level config.

Also, can't we do this?

```
export { recommended as default } from "@ilo-org/eslint-config";
```

config/eslint-config/README.md
```diff
@@ -10,42 +10,55 @@ npm install -D @ilo-org/eslint-config
...
     },
-  ],
-};
+  },
+];
 ```
 
+## Opinionated rules
+
+| Rule                                              | Status | Description                                                     |
```
bashlk, line 48: The status field of this table is a bit confusing. Does it mean that the config errors out for certain rules while warning about others?

A description about the status will be useful.

bashlk: In ESLint lingo, this is called severity - https://eslint.org/docs/latest/use/configure/rules#rule-severities

config/eslint-config/variants/base-ts.config.js
```diff
@@ -0,0 +1,59 @@
...
+    ],
+    // Migrated from old typescript-eslint config
+    "@typescript-eslint/ban-ts-comment": "off",
+    "@typescript-eslint/ban-ts-ignore": "off",
+  },
+};
+
+export default ts.config(
+  ...BaseJSConfig,
+  ...RecommendedTSConfig,
```
bashlk, line 52: Shouldn't the recommended overwrite the base?

packages/twig/cypress/e2e/navigation/navigation.cy.js
```diff
@@ -1,3 +1,5 @@
+/* @ESLINT-DEBT During new eslint refactoring this file was omitted because of heavy type refactoring */
```
bashlk, line 1: Ideally these ESLINT-DEBT annotations should reference the issue that describes why they were placed and how they should be fixed.

config/eslint-config/README.md
GGKapanadze: 1. Yes, project-specific rules, should be placed inside the local config, like the PR does for the `twig` package.
2. Since the mentioned config has general rules about `ts` and `js` files, and can be used for any project, I think the name `recommended` is better suited. IMHO by naming it `recommended` we encourage the users to extend it.

config/eslint-config/README.md
GGKapanadze: No, we don't have a named export.

config/eslint-config/variants/base-ts.config.js
GGKapanadze: 1. the target different file extensions
2. if the file extensions are the same and different rules are targeted they will be merged

packages/twig/cypress/e2e/navigation/navigation.cy.js
justintemps: @GGKapanadze to create the issue and then add the issue number to the comment.

Issue #1189, opened 2024-09-03
CSS variables support to dataurlicon


GGKapanadze:
# Problem

As mentioned in #1179, current implementation of `dataurlicon` and `colortodataurlicon` respectively doesn't support CSS variables

Issue #1080, opened 2024-06-24
Twig: improve doc page styles


justintemps:
Right now, the Doc page is using a combination of default Storybook styles and topography settings from the DS. The design is a mismash of type and spacing that doesn't look very nice. 

<img width="1290" alt="image" src="https://github.com/international-labour-organization/designsystem/assets/12401179/79d11b56-5448-4717-a7ba-d69f62966434">

Instead, we should be applying the same styles here that we use for the [RichText component ](https://twig.ui.ilo.org/?path=/docs/components-content-rich-text--rich-text)

We should also come up with nicer code formatting.

justintemps:
We also need to remove other SCSS variables from `_typography` and `_typographymonorepo` 

```scss
html,
body {
  color: $color-font-base; // <--- this
  font-family: var(--ilo-fonts-copy);
  font-size: #{$type-base-size}; // <--- this
  line-height: #{$type-base-line-height};
  letter-spacing: $type-base-letter-spacing; // <--- this
  -webkit-font-smoothing: antialiased;
}
```

PR #1233, opened 2024-09-25
Styles: Refactor richtext

### LINKED ISSUE #1079: Rich text: add line divider
We don't currently have an option to show a line divider in the rich text. Could this be added as per designs? See here: https://www.figma.com/design/qr8jQnxqJkRJFQnVC5PmrB/ILO-Foundations?node-id=6458-12248&t=VksMX48jFRmUZFfw-0 

Screenshot 2024-06-24 at 15 42 25

### LINKED ISSUE #1217: Docs(Twig): Replace placeholder image in RichText component story
It's missing

https://twig.ui.ilo.org/?path=/docs/components-content-rich-text--docs

image

justintemps:
## Description

This aims to simplify the Richtext stylesheets and harmonize the spacing with the designs in Figma. Also adds tests to make sure the component is respecting the spacing rules in the designs.

Fixes #1217 
Fixes #1079 

### Simplify the Richtext CSS

- [x]  Remove complicated map-get declarations
- [x]  Remove references to old theme package tokens

### New CSS variables in Foundations file

- [x]  font-size
- [x]  font-weight
- [x]  line-height
- [x]  letter-spacing

### Remove unused SASS mixin

`$font-weights` was removed from

- [x] packages/styles/scss/_typography.scss
- [x] packages/styles/scss/_typographymonorepo.scss

### Move blockquote css to a separate file

It now gets imported as a mixin. It was cluttering up the richtext stylesheets. This was we can use it for a standalone bockquote component down the line

### Add a utility class for responsive videos

`.responsive-video-embed` can be used for 16:9 video embeds (iframes) to make them responsive.

### Improve documentation

- [x] Explain how the component works
- [x] Add instructions for the utility class
- [x] Add instructions for the blockquote

### Add tests

- [x] Margin spacing between elements
- [x] Mobile styles 
- [x] Utility class for responsive embeds 

## Added bonus

This also fixes some issues with the Storybook styles in `packages/twig/.storybook/styles.scss` that were leaking into the  styles in the RichText component and probably other places too.


packages/styles/scss/components/_blockquote.scss
```diff
@@ -0,0 +1,112 @@
+@use "../tokens" as *;
+@use "../functions" as *;
+@import "../mixins";
+
+/**
+This mixin is used in _richtext.scss component to style blockquotes.
+It could eventually be used for a standalone blockquote component.
+**/
+@mixin blockquote {
```
GGKapanadze, line 9: If we are implementing the new styles I think it should be compliant to the specification we defined in #999

in this case, we should remove all SCSS variables (excluding `dataurlicon` #1189 ), define the blockquote variables, and reference them

packages/styles/scss/_typography.scss
```diff
@@ -37,32 +37,3 @@ html[lang*="ja"] {
   --ilo-fonts-display: var(--ilo-fonts-display-jp);
   --ilo-fonts-copy: var(--ilo-fonts-copy-jp);
 }
-
```
GGKapanadze, line 40: Nice, lets also remove other SCSS variables from `_typography` and `_typographymonorepo` 

```scss
html,
body {
  color: $color-font-base; // <--- this
  font-family: var(--ilo-fonts-copy);
  font-size: #{$type-base-size}; // <--- this
  line-height: #{$type-base-line-height};
  letter-spacing: $type-base-letter-spacing; // <--- this
  -webkit-font-smoothing: antialiased;
}
```

packages/styles/scss/components/_richtext.scss
```diff
@@ -1,68 +1,72 @@
 @use "../tokens" as *;
 @use "../functions" as *;
+@import "./blockquote";
 @import "../mixins";
 
 .ilo--richtext {
```
GGKapanadze, line 6: The same goes there, as mentioned in #999 if we want to expose anything to end user we should define them as a component-scoped variables, and let's remove `scss`  ones,  like `$color-font-caption`

GGKapanadze (changes requested):

packages/styles/scss/theme/_foundation.scss
```diff
@@ -87,6 +80,45 @@
...
+  --ilo-font-size-2xsm: #{px-to-rem(10)};
+  --ilo-font-size-xsm: #{px-to-rem(12)};
+  --ilo-font-size-sm: #{px-to-rem(14)};
+  --ilo-font-size-md: #{px-to-rem(16)};
+  --ilo-font-size-lg: #{px-to-rem(18)};
+  --ilo-font-size-xlg: #{px-to-rem(24)};
+  --ilo-font-size-2xlg: #{px-to-rem(28)};
+  --ilo-font-size-3xlg: #{px-to-rem(32)};
+  --ilo-font-size-4xlg: #{px-to-rem(36)};
+  --ilo-font-size-5xlg: #{px-to-rem(44)};
```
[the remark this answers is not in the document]
justintemps: @GGKapanadze I have a dumb question because I don't remember what we decided here. The scale value for the foundations file is 1. so we're just multiplying all these values by 1?

packages/styles/scss/components/_blockquote.scss
justintemps: Hey @GGKapanadze, we definitely should, I've just created issue for that https://github.com/international-labour-organization/designsystem/issues/1240

packages/styles/scss/_typography.scss
justintemps: I'll get to this in #1080

packages/styles/scss/components/_richtext.scss
justintemps: @GGKapanadze not sure what you're referring to here.

packages/styles/scss/theme/_foundation.scss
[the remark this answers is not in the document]
GGKapanadze: Yep, the initial value for the scale will be one but it will be exposed to anyone who wants to use the theme and can be easily scaled

packages/styles/scss/components/_richtext.scss
justintemps: Oh wait no I see it, good catch! @GGKapanadze

GGKapanadze: - There are still some variables defined as `SCSS` ones 
 - If we want expose any `richtext` scoped variables this is a good place 

but I think we can do it in scope of #1240

GGKapanadze (approved):

Issue #823, opened 2024-02-19
Twig: Implement Local Navigation


justintemps:
## ⚠️ This can't be started until #1302 has also been completed. Once it's finished, both can be merged to production

This ticket will implement in Twig the refactoring of the Navigation component and introduction of the separate Local Navigation component carried out in #1302. 

Note that #1302 may not include styles for the `complex` version of the Local Navigation component. If that's the case, then this ticket should add the complex version to both Twig markdown and styles, and then create a new ticket to add the complex version to React.

[See designs for further reference](https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?m=auto&node-id=3979-4509&t=JIzo17dLg17Px4ez-1)

bashlk:
@justintemps Do you have more specific suggestions for how this can be refactored? Perhaps split into two, similar to the navigation and local nav components?

justintemps:
@bashlk exactly, it should be refactored so there are two different scss files, one for each component.

justintemps:
@bashlk, @GGKapanadze I'm thinking this can wait for after the version 1. What do you think?

bashlk:
@justintemps I don't think we have worked a lot with the Navigation component so far so it is hard for us to weigh in on this. But since this is an internal detail of the design system, I think it can be postponed.

bashlk:
@justintemps We had a chat about this and we think that the Navigation component should probably be down in several separate components (e.g. language selector, search box, dropdown menu, horizontal nav). 

We also think that we should avoid having compound elements like the Navigation since they are not reusable and instead have simpler components like the ones listed above so that a consumer can combine them on their own.

justintemps:
Hey @ademarco and @GGKapanadze, just to make sure you're aware I'm working on this.

PR #1261, opened 2024-12-04
Localnav responsive menu

GGKapanadze:
# Description 

This PR fixes the responsive "burger menu" error that was introduced during `localnav` migration from `wingsuit` to `maestro`

https://github.com/user-attachments/assets/8f9339c0-8c69-4281-adeb-ba6c9b8f1576

## Future context

`navigation.scss` file contains a big overhead and we have an open issue about it #823, but the subject is a bit more complex. Despite `navigation.scss` being the complex implementation that is too big to be a "component" it also contains the styles for `localnav` and also `localnav` isn't working without the Behavior file from the `Navigation` component, so this 2 components are deeply intertwined, after finishing the #823 we should open the follow-up ticket for `localnav` separation

### DIFF
--- a/.changeset/gold-pillows-wait.md
+++ b/.changeset/gold-pillows-wait.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/twig": patch
+---
+
+fixed the localnav menu issue for responsive screens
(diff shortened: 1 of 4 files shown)


justintemps:
Thanks @GGKapanadze, when the menu is open it should show the light themed logo in the docs. If you could add that, then this is ready to merge.

GGKapanadze:
The logo is drawn by the `prop` named `{{mobilelogo.image}}` I'll check how it was passed previously

PR #1276, opened 2024-12-10
docs(twig): update Drupal-related documentation

ademarco:


justintemps:
Thanks for this @ademarco. Just FYI, this is only a docs change so we don't need a changeset for it.

PR #1285, opened 2024-12-13
Turbo repo upgrade

GGKapanadze:
# Description 

This PR upgrades the `turbo` version from `1.7.4` to `2.3.3`, but it also organizes and simplifies package scripts

## New turbo config 
We have a new `turbo.json` file that creates a correct dependency graph for packages with proper caching.

- This is a `build:libs` output during the first run,  without cache
```sh
Tasks:       9 successful, 9 total
Cached:      0 cached, 9 total
Time:        14.127s
```
- And here is a full cache version 
```sh
Tasks:     9 successful, 9 total
Cached:    9 cached, 9 total
Time:      493ms
```

Development scripts also got updated, `react:dev:docs` and `twig:dev:docs`  had no dependency tree, and during the first run you have to build all libraries and only then can you interact with `storybook`. New configuration includes an individual dependency list for both the `twig` and `react` package 

> Here is an example of a twig package

```json
"@ilo-org/react#dev:storybook": {
  "dependsOn": [
    "@ilo-org/styles#build:lib",
    "@ilo-org/icons#build:lib",
    "@ilo-org/brand-assets#build:lib",
    "@ilo-org/fonts#build:lib",
    "@ilo-org/utils#build:lib",
    "@ilo-org/themes#build:lib"
  ],
  "persistent": true,
  "cache": false
}
```

## Build script simplification

For the previous config to work, we had scripts like this 
```json
{
  "build": "pnpm run clean && node tasks/build.js",
  "build:lib": "pnpm build",
  "build:docs": "pnpm build"
}
``` 
The updated turbo implementation with a correct dependency tree eliminates the need to have duplicated build commands for each package and only keeps `build:lib` 

## Root scripts

The root script are all the same, only `clean:test` was removed.


turbo.json
```diff
@@ -38,6 +24,35 @@
...
+        "@ilo-org/icons#build:lib",
+        "@ilo-org/brand-assets#build:lib",
+        "@ilo-org/fonts#build:lib",
+        "@ilo-org/utils#build:lib",
+        "@ilo-org/themes#build:lib"
+      ],
+      "persistent": true,
+      "cache": false
+    },
+    "@ilo-org/react#dev:storybook": {
```
justintemps, line 44: Is this supposed to be `@ilo-org/react#storybook`?

justintemps (changes requested):
Hey @GGKapanadze, this is awesome, I think there's just a type in the turbo job for react. Running `pnpm react:dev:docs` returns this error:

```
turbo 2.3.3

  × missing tasks in project

Error:   × could not find task `@ilo-org/react#storybook` in project

 ELIFECYCLE  Command failed with exit code 1.
```

PR #1315, opened 2025-02-03
Fix(twig): data card size and columns.

### LINKED ISSUE #1270: Twig/React: Data Card narrow size layout with two columns is broken
Setting the `size` property to narrow should override the `columns` property and force there to be only one, because there's no design for a narrow card with two cols. 

## React

image

## Twig

image

justintemps:
Data Card should only render in two columns if `size` is set to `narrow` or `wide`

Fixes #1270

### DIFF
--- a/packages/twig/src/components/card_data/card_data.twig
+++ b/packages/twig/src/components/card_data/card_data.twig
@@ -1,6 +1,12 @@
 {# card_data.twig #}
 
+{# Defaults to one column #}
 {% if not columns %}
+	{% set columns = "one" %}
+{% endif %}
+
+{# size overrides column if it's set to 'narrow' #}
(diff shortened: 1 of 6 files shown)


packages/twig/src/components/card_data/card_data.twig
```diff
@@ -1,9 +1,15 @@
 {# card_data.twig #}
 
+{# Defaults to one column #}
 {% if not columns %}
 	{% set columns = "one" %}
 {% endif %}
```
justintemps, line 7: Hey @GGKapanadze, minor change to note for the Data Card. See below. Basically `size` should override `columns` if `size` is set to `narrow`.

GGKapanadze (approved):

PR #1320, opened 2025-02-05
React: add standalone blockquote component

justintemps:
Implements the Blockquote comonent in React.

Only merge this after #1319

### DIFF
--- a/packages/react/src/components/Blockquote/Blockquote.props.ts
+++ b/packages/react/src/components/Blockquote/Blockquote.props.ts
@@ -1,4 +1,0 @@
-export interface BlockquoteProps {
-  quote: string;
-  cite: string;
-}
--- a/packages/react/src/components/Blockquote/Blockquote.props.ts
+++ b/packages/react/src/components/Blockquote/Blockquote.props.ts
@@ -1,0 +1,4 @@
+export interface BlockquoteProps {
+  quote: string;
+  cite: string;
+}
--- a/packages/react/src/components/Blockquote/Blockquote.tsx
+++ b/packages/react/src/components/Blockquote/Blockquote.tsx
@@ -1,16 +1,22 @@
-import React from "react";
-import { BlockquoteProps } from "./Blockquote.props";
+import { forwardRef } from "react";
 import useGlobalSettings from "../../hooks/useGlobalSettings";
 
-const Blockquote: React.FC<BlockquoteProps> = ({ quote, cite }) => {
-  const { prefix } = useGlobalSettings();
+export interface BlockquoteProps {
+  quote: string;
+  cite: string;
+}
 
-  return (
-    <blockquote className={`${prefix}--blockquote`}>
-      <p>{quote}</p>
-      <cite>{cite}</cite>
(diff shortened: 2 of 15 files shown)


packages/react/src/components/Blockquote/Blockquote.props.ts
```diff
@@ -0,0 +1,4 @@
+export interface BlockquoteProps {
+  quote: string;
+  cite: string;
+}
```
GGKapanadze, line 4: There is no need to keep the props in a separate file. I tried to move them inside the component. For the feature, when we relocate the story file to the component directory, we can introduce a new file tree. Until then, let's keep the props inside the component. 

And let's use types instead of interfaces, we can better compose them

packages/react/src/components/Blockquote/Blockquote.tsx
```diff
@@ -0,0 +1,16 @@
...
+const Blockquote: React.FC<BlockquoteProps> = ({ quote, cite }) => {
+  const { prefix } = useGlobalSettings();
+
+  return (
+    <blockquote className={`${prefix}--blockquote`}>
+      <p>{quote}</p>
+      <cite>{cite}</cite>
+    </blockquote>
+  );
+};
```
GGKapanadze, line 14: We need to expose the ref for all the components

packages/react/src/components/Blockquote/Blockquote.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default Blockquote;
```
GGKapanadze, line 16: And let's stick to the named exports, the default exports can't be type augmented 

https://github.com/Microsoft/TypeScript/issues/14080

packages/react/src/stories/Blockquote/Blockquote.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+            a citation.
+          </Description>
+          <Primary />
+          <Heading>Props</Heading>
+          <ArgTypes of={BlockquoteMeta} />
+          <Stories />
+        </>
+      ),
+    },
+  },
```
GGKapanadze, line 49: This part isn't necessary if we don't do the custom story docs, you can add a description like this 

````tsx
const BlockquoteMeta: Meta<typeof Blockquote> = {
  title: "Components/Content/Blockquote",
  component: Blockquote,
  tags: ["autodocs"],
  argTypes: {
    quote: {
      description: "The quote text",
      control: "text",
      table: {
        type: { summary: "string" },
      },
    },
    cite: {
      description: "The citation for the quote",
      control: "text",
      table: {
        type: { summary: "string" },
      },
    },
  },
  parameters: {
    docs: {
        description: " The Blockquote component is used to display a quoted text along with a citation"
    },
  },
````

GGKapanadze (changes requested):

justintemps:
Hey @GGKapanadze thanks for the review, would you mind having a last look before I merge?

GGKapanadze (approved):

PR #1337, opened 2025-02-12
Language Toggler 

GGKapanadze:
# Description 

This PR introduces a new `LanguageToggler` component. It's a generic implementation that can be used across all three implementations right now 

https://github.com/user-attachments/assets/9f5c34e4-db20-4966-9b34-2cab1f1c83b3

## API

The new codewise components look like this

```jsx
<LanguageToggle
  language="English"
  options={[
    {
      label: 'Spanish',
      url: '#'
    },
    {
      label: 'German',
      url: '#'
    },
    {
      label: 'French',
      url: '#'
    }
  ]}
/>
```

- It respects the theme
- It allows you to hide the `globe` icon, with the `hideIcon` prop
- It uses the native context menu as the breadcrumb
- It allows you to pass all props to the native `div` element 

## Styles

The styles are done with the #999 rules. It utilizes the CSS variables + new foundation and new `icon` mixin. Background color is implemented via component variables so it can be easily manipulated by the parent element(`Navigation`)

### DIFF
--- a/.changeset/red-fishes-kiss.md
+++ b/.changeset/red-fishes-kiss.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/react": minor
+---
+
+Implemented a new LanguageToggler component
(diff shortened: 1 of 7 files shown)


GGKapanadze:
@justintemps After merge we should open the follow up ticket to implement the same component for the twig

justintemps (approved):

justintemps:
hey @GGKapanadze this looks good, you can go ahead and merge.

PR #1377, opened 2025-03-17
Canary Batch 

GGKapanadze:
# Canary Batch 

This PR introduces the changes done in the canary branch for the `@ilo-org/react` and `@ilo-org/styles` packages.

## Changes

- Header component now renders portal components only on the client
- Breadcrumbs got the support for the external Link component
- `useLayoutEffects` was switched to `useEffect` in  favor of rendering in next
- `isActive` now supports callback
- SSR check for `useBreadkpoint`
- Mobile navigation doesn't renders the `more` button if there are not enough menu items
- Context menu now has an initial positioning
- `flex-start` was instead of `start`


justintemps:
Hey @GGKapanadze, these are all fine but it seems like we should not be merging changesets from `canary` to `develop` if I've understood [this advice correctly](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md#what-to-do-with-the-snapshot-branch)

PR #1378, opened 2025-03-17
Canary Release Workflow

GGKapanadze:
# Canary Release Workflow

I initially tried to solve our dependency issues using GitHub linking as we discussed, which would have been a simpler approach. However, this didn't address all our needs, particularly with how internal package dependencies are structured.

1. First I tried to link dependencies via GitHub using the `yarn` but since we have a mono-repo to correctly resolve the packages we need to only pull `subdirectory` of our repo, which `yarn` isn't capable of 
2. I tried to move to `pnpm@10`  since it supports pulling the GitHub branch with a subdirectory like this 
`github:international-labour-organization/designsystem#canary-with-pck-fix&path:/packages/react` but we have workspace level dependencies `("@ilo-org/styles": "workspace:*")` and `pnpm` was unable to resolve those
3. There are a few hacks  like patching to package JSON or removing the dependencies but those will introduce extra complexity and confusion 

So `changeset` has a [snapshot](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md) release concept exactly for this use case, so this PR will tag every commit on a canary as a `canary-{COMMIT_HASH}`.

The contracts remain the same, once I open the PR for the ilo live I'll also create a PR for `canary`

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -1,0 +1,71 @@
+name: Canary Release
+
+on:
+  push:
+    branches:
+      - canary
+
+env:
+  CI: true
+  PNPM_CACHE_FOLDER: .pnpm-store
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release-canary:
+    name: Canary Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


justintemps:
Hey @GGKapanadze, I agree with the approach but I think we need to carefully document the workflow.

From what I can understand in the docs:

1. We should continue to develop features on feature branches
2. To test those features, feature branches can be merged into the canary branch which will generate canary releases. 
3. Changesets should not be added to canary releases. 
4. We should not merge from `canary` into `develop` or `main`, but only from `feature` to `canary` or `feature` to `develop` to `main`.

Is this correct?

PR #1392, opened 2025-04-02
Canary Batch

GGKapanadze:
# Canary Batch

This PR introduces the changes done in the canary branch for the @ilo-org/react and @ilo-org/styles packages.

## Changes
- corrected typography for the `ScoreCard`
- `Tabs` can now accept a default value for the opened tab 
- `MultiLinkCard` now can accept intro as a `ReactNode`
- `ScoreCard` can receive `className` and content as a `ReactNode` too

### DIFF
--- a/.changeset/brave-singers-vanish.md
+++ b/.changeset/brave-singers-vanish.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/react": patch
+---
+
+default value for the active tab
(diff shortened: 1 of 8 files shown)


justintemps:
Hey @GGKapanadze this all looks good. 

I've developed a personal convention for changesets like this:

**Component:** Description of the change

Changes are usually specific to components and I think this makes it easier for library users to understand where to look for changes up front. What do you think?

PR #1421, opened 2025-04-22
Feat(Twig): StatefulComponent

justintemps:
## Introduction

This PR proposes a utility class to the Twig package called `StatefulComponent`. It provides simple reactive state system that triggers handlers automatically when state changes.

## Background

While working on #823, I noticed that the Navigation component has to manage a bunch of state (whether the mobile drawers are open are closed in desktop or mobile where there are more than one). There are several others components like LanguageToggle and Breadcrumb that have to manage state as well, albeit it a little less. 

We have tickets open for new components like a photogallery that might have to manage a lot more state.

At the same time, I really like that the Twig package is lightweight in terms of JavaScript. We've done a good job of "just using the platform" wherever possible and I think we should continue doing so. Fortunately, modern JavaScript provides all the tools we need to manage state in Twig components in a straightforward, declarative way without having to reach for libraries. 

## StatefulComponent

This is a superclass that can be extended to add statement management functionality as needed. It uses a simple object Proxy to track state and fire callbacks when state changes. That's all there is to it.

## Usage

This is a superclass that can add state management to components as needed. Here's an example:

```js
class MyComponent extends StatefulComponent {
 constructor(element) {
   // Initialize state with default values
   const initialState = {
     isOpen: false,
     selectedItem: null,
     items: []
   };
 
   // Call super with element and initial state
   super(element, initialState);
 
   // Initialize component-specific properties
   this.prefix = 'my-component';
 
   // Initialize the component
   this.init();
 }
 
 init() {
   this.cacheDomReferences()
     .bindHandlers()
     .registerStateHandlers();
   return this;
 }
 
 registerStateHandlers() {
   // Register handlers for state changes
   this.registerStateHandler('isOpen', (value) => {
     if (value) {
       this.open();
     } else {
       this.close();
     }
   });
 
   this.registerStateHandler('selectedItem', (value) => {
     this.updateSelectedItem(value);
   });
 
   return this;
 }
 
 // Example method that updates state
 toggle() {
   this.state.isOpen = !this.state.isOpen;
 }
}
```

All we have to do is initialize state in the constructor and from then on, we can register event handlers that will file whenever that state changes. From then on, things like click handlers can simply set state as needed with a simple assignment and the stateHandlers will do the rest. Best of all, you can easily use state in checks without having to resort to hacks like checking the DOM or looking at data attributes.

## ⚠️ Caveat

As of the time of the first pull request, this will only work for direct assignments! For example, this will fire an update:

```js
this.state.myValue = true
```

But this won't 😒 

```js
this.state.items.push("item-1");
```

We don't have a use case for the second scenario YET but I think we probably will. And anyway, if you've worked with any framework or state management lib at all, you would expect the second option to work.

This can be done by recursively adding Proxies to objects passed as values, which I think is worth the effort. I'm going to research a little more and then add.

### DIFF
--- a/packages/twig/cypress/e2e/utils/statefulComponent.cy.js
+++ b/packages/twig/cypress/e2e/utils/statefulComponent.cy.js
@@ -1,0 +1,189 @@
+import { StatefulComponent } from "../../../src/utils/statefulComponent";
+
+describe("StatefulComponent", () => {
+  let component;
+  let element;
+  let initialState;
+  let stateChangeHandler;
+
+  beforeEach(() => {
+    // Create a mock DOM element
+    element = document.createElement("div");
+    element.id = "test-component";
+    document.body.appendChild(element);
+
+    // Setup initial state
+    initialState = {
+      isOpen: false,
+      items: [],
+      nested: {
+        value: "initial",
+        count: 0,
+      },
+    };
+
+    // Create a spy for state change handler
+    stateChangeHandler = cy.spy();
+
+    // Initialize component
+    component = new StatefulComponent(element, initialState);
+  });
+
+  afterEach(() => {
+    // Clean up DOM
+    element.remove();
+  });
+
+  it("should initialize with correct state", () => {
+    expect(component.state).to.deep.equal(initialState);
+  });
+
+  it("should trigger handler when state property changes", () => {
+    component.registerStateHandler("isOpen", stateChangeHandler);
+    component.state.isOpen = true;
+    expect(stateChangeHandler).to.have.been.calledWith(true, "isOpen");
+  });
+
+  it("should not trigger handler when state property is set to same value", () => {
+    component.registerStateHandler("isOpen", stateChangeHandler);
+    component.state.isOpen = false; // Same as initial value
+    expect(stateChangeHandler).not.to.have.been.called;
+  });
+
+  it("should handle multiple handlers for same property", () => {
+    const handler1 = cy.spy();
+    const handler2 = cy.spy();
+
+    component.registerStateHandler("isOpen", handler1);
+    component.registerStateHandler("isOpen", handler2);
+
+    component.state.isOpen = true;
+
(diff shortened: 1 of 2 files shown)


justintemps:
Hey @GGKapanadze let me know if you wanted to offer some feedback on this before I go ahead and add support for objects/arrays as first-level state properties.

justintemps:
Hey @bashlk yea, I think we have two options for that:

1. Do a shallow equality check with no library :

```this.state.obj.property = "value"``` 

will work but 

```this.state.nested.obj.property = "value"``` 

won't.

This is my preferred option right now. I've never thought it was a good idea to pass around nested objects or multidimensional arrays as state.  There's always a price to pay for that, both in terms of performance and complexity, and its an open invitation for the developer to do bad things.

2. Use a library to do a deep check

The one you mentioned has been archived, but I found [js-proxy-deep](https://github.com/samvv/js-proxy-deep) which looks pretty comparable. It's relatively small and looks pretty evergreen, though it would still add more complexity than I think is warranted.

99% of state in the Twig library are boolean checks. More complicated components like a Photo Gallery could eventually require more than that, but adding support for arrays and objects with a shallow equality check should be enough.

What do you think @bashlk and @GGKapanadze ?

**Comment by justintemps** [comment]
Hey @bashlk and @GGKapanadze, so I've gone down a hole rabbit hole on this one and have finally decided that it's a problem that's not worth solving. 

For the purposes of Twig components, we can simply say that state works via assignment only.

E.g.

```
this.state.myArray = [...this.state.myArray, "newItem"]; 
```

will fire the stateHandler, but 

```
this.state.myArray.push("newItem");
```

won't. 

I can live with that, what do you think?

justintemps:
Just to add, that the above allows us to keep the current implementation with no changes and no libraries. I think this is especially important given how the package is used by Drupal. Essentially, any utility or library we import into the class file is going to get added once for every behaviours file it's used in. There's no module federation setup at all yet. 

That means that if you a library used in two different components and they're both on the page, then the library is going to get loaded twice.

Issue #1423, opened 2025-04-25
React: improve accessibility of SubsiteNav


justintemps:
The React version of this component is missing some important accessibility features:

- [ ]  Clicking the more button should focus the first element of the dropdown
- [ ] The dropdown should use a focus trap to improve tab navigation
- [ ] The LanguageToggle should also focus the first element of the dropdown when clicked
- [ ] The LanguageToggle should also employ a focus trap
- [ ] In mobile, touching the hamburger menu should give focus to the first focusable element of the main drawer
- [ ] In mobile, touchingthe Language and More buttons should give focus to the first focusable element of the secondary drawers

Things to check:

- [ ] The More Button, LanguageToggle button and Hamburger menu all have the necessary aria attributes set when they're respective menus are open
- [ ] These buttons should also have `aria-controls` attributes pointing to the menu ids.
- [ ] None of the drawers or menus should be accessible to tab navigation when they're hidden

PR #1424, opened 2025-04-25
Implement SubsiteNav for Twig

### LINKED ISSUE #823: Twig: Implement Local Navigation

### LINKED ISSUE #1422: Bring SubsiteNav into line with updated Figma designs
The SubsiteNav doesn't comply with the [latest designs in Figma](https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?m=auto&node-id=3979-4509&t=4cb9h0M5uXBEolet-1). 

It needs to be updated before it can be used in production.

justintemps:
## Introduction

This PR implements the `SubsiteNav` component for Twig. It follows the same patterns as the React version, with only slight adjustments to the component’s styles. Fixes #823.

## Deploy Preview

🫣 [See the Subsite Nav here](https://deploy-preview-1424--ilo-ui-twig.netlify.app/iframe.html?args=&id=components-navigation-subsitenav--default)

## Next steps

Before we can release this production, we need to tackle the following issues:

- #1421 will finalize the `StatefulComponent`, introducing a standard way of handling state management in Twig components; this pattern is used extensively in the `SubsiteNav`.
- #1422 will align the component with the final Figma designs.

## Accessibility

This version of the component differs from the React version only in its improved accessibility. It ensures that all dropdown menus and mobile drawers are accessible and manages tab navigation via focus traps. Issue #1423 has been opened to address this.

## New utilities

Implementing this component required adding three new utilities to the Twig package. These utilities are specific to our Twig implementation and aren’t intended for use in other packages.

### `StatefulComponent`

This PR makes extensive use of a new `StatefulComponent` class (introduced in #1421). It’s an extensible JavaScript class that provides simple state management. In this PR, it supports only basic assignments of `string` and `boolean` values.

### `createFocusTrap`

This utility creates a focus trap for Tab navigation that can be exited by pressing the `Escape` key. It takes four arguments:

1. A keyboard event  
2. A list of focusable elements  
3. A callback to fire when the `Escape` key is pressed  
4. An optional callback to fire when the `Tab` key is pressed  

```js
createFocusTrap(
  event,
  this.dropdown.querySelectorAll("a"),
  () => {
    this.state.dropDownIsOpen = false;
    this.dropdownButton.focus();
  },
  null
);
```

### `createBreakpointObserver`

This utility monitors the window size using `ResizeObserver` and invokes callbacks when the breakpoint changes:

```js
this.breakpointObserver = createBreakpointObserver((breakpoint) => {
  this.state.isDesktop = ["xl", "xxl"].includes(breakpoint);
});
```

## Dependencies

- **Cypress** upgraded to **v14.3.2**  
  - Leverages the new tab-navigation testing methods introduced in this release, which are required for our tests.


packages/styles/scss/components/navigation/_nav-menu.scss
```diff
@@ -14,16 +15,18 @@
 
   &__list {
     display: flex;
-    gap: spacing(12);
+    // gap: spacing(12);
```
bashlk, line 18: Commented out code still present.

packages/styles/scss/components/navigation/_nav-menu.scss
```diff
@@ -14,16 +15,18 @@
...
     align-items: center;
+    height: 100%;
 
     @include breakpoint("xl", true) {
-      gap: spacing(6);
+      // gap: spacing(6);
```
bashlk, line 23: Same comment as above

packages/twig/cypress/e2e/navigation/nav_mobile.cy.js
```diff
@@ -0,0 +1,168 @@
...
+        cy.get(".ilo--nav-mobile-menu__link").should(
+          "have.length",
+          fixture.facadeItems.length
+        );
+      });
+    });
+
+    it("the currently selected menu item includes a marked icon", () => {
+      cy.get("[class$='burger']").click();
+      cy.get("#ilo--nav-mobile-drawer__main").within(() => {
```
bashlk, line 52: The outer within could probably be removed

packages/twig/rollup.config.js
```diff
@@ -58,11 +58,15 @@ const copyConfig = copy({
       },
     },
     {
```
bashlk, line 60: Why was the rollup config changed?

packages/twig/src/components/nav/nav.js
```diff
@@ -0,0 +1,291 @@
...
+
+    // Prefix
+    this.prefix = this.element.dataset.prefix;
+
+    // References to elements that get rendered dynamically on the client
+    this.dropdown = null;
+
+    // Set up a breakpoint observer to track viewport size changes
+    this.breakpointObserver = createBreakpointObserver((breakpoint) => {
+      this.state.isDesktop = ["xl", "xxl"].includes(breakpoint);
```
bashlk, line 43: It seems like this logic is repeated in several places. Maybe createBreakpointObserver can be modified to only trigger when entering the given breakpoints.

packages/twig/src/components/nav/nav.js
```diff
@@ -0,0 +1,291 @@
...
+
+  /**
+   * Initializes the component by setting up all necessary functionality.
+   * This includes rendering client-side content, caching DOM references,
+   * binding event handlers, registering state handlers and starting the breakpoint observer .
+   *
+   * @returns {Nav} Returns the instance for method chaining
+   */
+  init() {
+    this.renderClientContent()
```
bashlk, line 65: Isn't this the same init as the Stateful component? Can't this be specified on top of super?

packages/twig/src/utils/breakpoint.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * Breakpoint utility that monitors DOM changes and fires callbacks
+ * when crossing breakpoint thresholds.
+ */
+
+// Define breakpoints matching the SCSS breakpoints
+const BREAKPOINTS = {
```
bashlk, line 7: Ideally we should find a way to use the breakpoint values set in SCSS without redefining them here. Once we redefine them here, then this needs to be kept in sync with the values set in the SCSS.

Failing this, I think the at least the specific file in which these files are set should be mentioned in the comment.

packages/twig/src/components/nav/nav.js
bashlk: If not, I think it would still be better if the stateful component took care of its own initialization so that users of it doesn't have to think about it. Just extend the class and start working directly with this.state.

packages/twig/src/components/nav/nav.js
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  /**
+   * Binds all event handler methods to the component instance.
+   * This ensures that 'this' context is preserved when handlers are called.
+   *
+   * @returns {Nav} Returns the instance for method chaining
+   */
+  bindHandlers() {
+    this.handleBreakpointChange = this.handleBreakpointChange.bind(this);
```
bashlk, line 131: We can skipt this manual binding by using the arrow function syntax. (e.g. `handleBreakPointChange => {}`)

packages/twig/rollup.config.js
justintemps: @bashlk Before the rollup config expected all of the .twig files to be at the top level of the component. The Nav however has .twig files in subfolders, so I had to update the config to make sure they got copied over appropriately.

packages/twig/src/components/nav/nav.js
justintemps: @bashlk repeated in which places? Anyway this is pretty much the same way the React component works using the "use-breakpoint" lib

packages/twig/src/components/nav/nav.js
justintemps: Hey @bashlk, no StatefulComponent has its own private `init` method that does different things than the `init` method in the component classes. They're not related at all.

packages/twig/src/utils/breakpoint.js
justintemps: @bashlk agreed, for the time being they're hardcoded in React too, but this is definitely an improvement we should make.

packages/twig/src/components/nav/nav.js
justintemps: @bashlk excellent point

packages/twig/src/components/nav/nav.js
bashlk: I mean this expression `["xl", "xxl"].includes(breakpoint)`. It's a minor nitpick but if the utility accepted the array of watched breakpoints directly, it can expose a callback that will return the boolean state of when those breakpoints are active or not.

justintemps:
Ok but what if you wanted to handle different breakpoints differently? This
just seems really flexible, you give it a callback that gets called every
time the breakpoint changes with the new breakpoint. The extra boilerplate
doesn’t seem like it warrants sacrificing the flexibility.

packages/twig/src/components/nav/nav.js
justintemps: Ok but what if you wanted to handle different breakpoints differently? This just seems really flexible, you give it a callback that gets called every time the breakpoint changes with the new breakpoint. The extra boilerplate doesn’t seem like it warrants sacrificing the flexibility.

PR #1440, opened 2025-05-09
Use static image only for SubsiteNav Logo

justintemps:
## Background

The SubsiteNav originally used an ordinary ILO Logo and added the sub-brand as an h3 attached to it.

image

This works fine when the sub-brand is nice and short, but unfortunately, the SubsiteNav will be used by websites that don't have succinct sub0-brands.

Here's an example of how a longer sub-brand name would break the compact layout of the SubsiteNav:

image

As a result, I think it's better just to use plain logo images that can't break. Yes, we'll need a way of standardising how these logos look and are created. I have another idea for that which involves using the [React Logo Component](https://react.ui.ilo.org/?path=/story/components-media-logo--sub-brand-logo) to generate downloadable logo images. This component was created to handle unwieldy sub-brand names and dynamically resizes the font of the sub-brand as it gets longer and longer.

This is not a breaking change to the current implementation in React, it will simply require us to use a different logo image with the sub-brand baked in.

## To merge

This branches off the `feat/twig/new-navigation` branch. We can merge that one first and then set this to merge into `develop`. Here I made the PR back to the feature branch so it would be easier to see what I changed.

### Complex Desktop 
image

### Compact Desktop
image

### Mobile Drawer
image


GGKapanadze (approved):
Let's merge it to the canary first to testrun for ilo live

PR #1447, opened 2025-05-14
Twig: Nav logo can have an optional link

### LINKED ISSUE #1446: Twig: New Subsite Nav logo is missing link
The SubsiteNav Logo right now is missing a link. This needs to be added and we probably also have to add a field so that users can determine what the link should be.

justintemps:
Fixes #1446


justintemps:
@ademarco it looks like tests here are failing because of https://github.com/international-labour-organization/ilo_base_theme/issues/72

Issue #1543, opened 2025-09-02
Image credit is not displayed correctly


bashlk:
Currently the image credit is rendered like this on lower resolutions.

<img width="563" height="518" alt="Image" src="https://github.com/user-attachments/assets/6fde5bd1-ddb5-43bb-8bf9-e3cdf7d761cf" />

Moreover, the corner cut that is shown on the designs is not present when the image credit is shown on higher resolutions.

<img width="952" height="677" alt="Image" src="https://github.com/user-attachments/assets/58afcd6b-d7bc-49f6-8896-8e8c4fdb69a1" />

These issues need to be fixed.

PR #1545, opened 2025-09-02
fix(styles): implemented mixins for image styles

### LINKED ISSUE #1241: Styles: use mixin for Image styles in Image and RichText component
Right now, we're basically copying the Image styles in the RichText component to style its images. 

You can see this done here: https://github.com/international-labour-organization/designsystem/pull/1233/commits/a9a1b7757d1c2266fda6bfb7a64710ece53a2049

This is not a good solution, instead we should use a mixin and import it in both places, the way we do with `_breadcrumb.scss` in `_richtext.scss`. That way, the styled will be kept in sync in both places.

bashlk:
Fixes #1241

This PR introduces several mixins for sharing image styles across the codebase. In addition to the RichText component described in the issue, I also found the caption styles used in the Video component, so I used the mixin there as well.

The Image component lacked support for the dark theme but the image in RichText had support for it, so I also added support for the dark theme to the Image component and updated it to match the designs. The image credit does not display correctly in dark theme and that will be tackled in #1543

### DIFF
--- a/.changeset/witty-kiwis-wonder.md
+++ b/.changeset/witty-kiwis-wonder.md
@@ -1,0 +1,8 @@
+---
+"@ilo-org/styles": patch
+---
+
+Implemented mixins for images and image captions and synced components with design.
+Image: synced image caption with design
+RichText: synced image caption with design
+Video: synced image caption with design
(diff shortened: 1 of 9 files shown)


justintemps (changes requested):
Hey @bashlk this looks good, you just need to add a changeset

justintemps (approved):

PR #1629, opened 2025-09-22
 Detail Card RTL Arrow

### LINKED ISSUE #1461: Detail card dateExtra arrow points the wrong way in RTL
Image

GGKapanadze:
# Description 

This PR fixes the `arrow` icon issue for the `Detail Card` component for both `react` and `twig` implementations

> React

https://github.com/user-attachments/assets/f396bf3c-7dac-42e7-bfc3-8398faeb7c09

> Twig

https://github.com/user-attachments/assets/b10fa0ad-d5eb-4f95-adca-da62f8680943

# Implementation

- The implementation is declaratively renders both `right` and `left` icons directly from the component(for both) and then styles are manipulated based on the direction.

- `triangle_left` is added to the icon pack

### DIFF
--- a/.changeset/moody-pugs-talk.md
+++ b/.changeset/moody-pugs-talk.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/icons": patch
+---
+
+New `triangle_left` icon added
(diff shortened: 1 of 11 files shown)


justintemps (changes requested):
Hey @GGKapanadze, all looks good, but I think adding a new icon to the icons package should always be a minor bump instead of a patch.

bashlk (approved):

justintemps (approved):

PR #1631, opened 2025-09-23
fix(react): fixed Dropdown value not updating after intial render

### LINKED ISSUE #1469: React: Dropdown doesn't update as value is changed
The Dropdown component doesn't update when the value prop is changed after the initial render. This is required when the component is used in a controlled fashion, where the current value is stored in state outside the component.

bashlk:
Fixes #1469

### DIFF
--- a/.changeset/spicy-numbers-burn.md
+++ b/.changeset/spicy-numbers-burn.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/react": patch
+---
+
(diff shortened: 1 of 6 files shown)


justintemps (changes requested):
@bashlk can we add a test for this?

GGKapanadze (approved):

justintemps (approved):

PR #1646, opened 2025-09-30
feat(react): implemented loading skeletons for card components

### LINKED ISSUE #1538: Create skeleton loader for cards
This is specifically needed to unblock this issue on the ILO Live Repository: https://github.com/international-labour-organization/ilo-live-frontend/issues/1184

Here are the designs: https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/branch/0C2b7P2nMu3uyv8NqTLr3x/ILO-Components-Library?node-id=3587-2231&t=K51CvoxyDxmc6Iej-0

See the parent issue for links to designs etc.

bashlk:
Fixes #1538 

This PR implements the loading skeletons for the card components in React. Since the design for the skeletons in the dark theme is not finalized, only the light theme is implemented for now. This PR also implements basic tests for the card components that were altered, to test the basic rendering and the loading skeleton.

### DIFF
--- a/.changeset/lazy-cherries-clean.md
+++ b/.changeset/lazy-cherries-clean.md
@@ -1,0 +1,15 @@
+---
+"@ilo-org/react": minor
+"@ilo-org/styles": patch
+---
+
+Implemented loading skeleton for card components through the `loading` prop.
+**FeatureCard:** added loading skeleton
+**TextCard:** added loading skeleton
+**DetailCard:** added loading skeleton
+**PromoCard:** added loading skeleton
+**MultiLinkCard:** added loading skeleton
+**DataCard:** added loading skeleton
+**FactsListCard:** added loading skeleton
+**StatsCard:** added loading skeleton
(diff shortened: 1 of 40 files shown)


bashlk:
@justintemps I think having a separate `CardSkeleton` component is a much better way to implement the skeletons. Ideally, design can maybe come up with a few variants that match common configurations of cards and then we can use the closest variant in the places that need a skeleton.

Since we don't have many client rendered pages, it is maybe better to keep the `CardSkeleton` as a completely separate component and use it only in places that need it.

justintemps:
thanks @bashlk, we're saying the exact same thing. I don't think design needs to do anymore work, it looks to me like the skeleton implementations you've added are already particular to each card. Let's just extract that from the card components themselves and move them into a dedicated CardSkeleton component.

bashlk:
Agreed to create seperate sub components for each card skeleton alongside the card components. Closing this as most of this will not be used anymore.

PR #1647, opened 2025-09-30
React: Main Nav component

GGKapanadze:
# Description 

This PR introduces a new navigation component, named `MainNav` that replaces the generic `Navigation`  one

https://github.com/user-attachments/assets/af09fb2f-80f5-421d-a4f1-a8ebbe0b7a2d

https://github.com/user-attachments/assets/a78f43c4-9ec8-437a-99fb-b54dd65ce21e

## New Search

The `MainNav` involves a new search type, that can be easy adjusted to all other navigation components.

Here is an interface 
```ts
type InputSearchProps = {
  type: "input";
} & (
  | { field: SearchFieldProps; component?: never }
  | { field?: never; component: ReactNode }
);
```

There are  two options, either you pass props for the `SearchField`  or you pass you `ReactNode` manually to render out your desired structure with your own control

## Navigation Dropdown Shuffling 

The Implementation also handles overlap prevention when you need to switch between two `NavigationDrodown` s

https://github.com/user-attachments/assets/c5a83297-8dad-45fc-8009-28a8d84c557b

## useClickOutside
 
New hook was introduces to handle the outside clicking, with exception supports(after merging this PR I will open follow up ticket to add this across all navigations and for components like, `contextmenu` or `breadcrumbs`)

```ts
  useOutsideClick(searchButtonRef, () => {
    if (isSearchOpen) {
      setIsSearchOpen(false);
    }
  }, [searchDropdownRef]);
```

##  File structure

File structure was simplified for both `react` components and `style` files 

image

image

### DIFF
--- a/packages/styles/scss/components/navigation/index.scss
+++ b/packages/styles/scss/components/navigation/index.scss
@@ -1,4 +1,4 @@
+@use "internal";
 @use "nav-compact";
 @use "nav-complex";
 @use "nav-main";
-@use "internal";


packages/react/src/components/Nav/internals/NavigationDropdown.tsx
```diff
@@ -22,16 +30,16 @@ type NavigationDropdownProps = {
...
-}: NavigationDropdownProps) => {
+const NavigationDropdownBare = forwardRef<
+  HTMLDivElement,
+  NavigationDropdownProps
+>(({ isOpen, className, navRef, children }, ref) => {
   const { prefix } = useGlobalSettings();
 
   const wrapperRef = useRef<HTMLDivElement>(null);
 
+  useImperativeHandle(ref, () => wrapperRef.current!, []);
```
bashlk, line 41: Why is this necessary?

packages/react/src/hooks/useOutsideClick.ts
```diff
@@ -0,0 +1,32 @@
+import { useEffect } from "react";
+
+function useOutsideClick<T extends HTMLElement>(
+  ref: React.RefObject<T>,
```
bashlk, line 4: You could combine `ref` and `exceptions` into a single parameter called `elements` or `inside` for example because you are detecting for clicks outside of both.

bashlk:
I am lacking the context for why this change is required but the new component is still quite hard to follow. I think trying to share things between the two components is making things more complicated than it needs to be.

packages/react/src/components/Nav/MainNav.tsx
bashlk, line 1: We should probably start the practice of writing tests for any components that we add / modify.

PR #1651, opened 2025-10-06
Twig: main-nav component

GGKapanadze:
# Description 

⚠️  Merge after #1647 

This PR introduces a new  main-nav component for the `twig` design system

https://github.com/user-attachments/assets/8615189b-a48e-4d8e-929d-3d14070dc098

https://github.com/user-attachments/assets/87106eff-be60-4d6f-bbe6-93cc1269a6a5

## Search

New search is implemented inside of `navsearch.twig` and JS bundle is reworked to allow use of both drop-downs, double open collision  is also handled. Search functionality is handled with the same `form` component so migration should not cause many problems, search fields also have dedicated selector to target them with ease 

## Styling

All styles are shared, only 4 lines of css is dedicated to twig to handle the form sizing

### DIFF
--- a/packages/twig/src/components/nav/desktop/navsearch.twig
+++ b/packages/twig/src/components/nav/desktop/navsearch.twig
@@ -1,0 +1,25 @@
+{# navsearch.twig #}
+{% set prefix = prefix|default("ilo") %}
+{% set base_class = prefix ~ "--nav-dropdown" %}
+
+<template id="{{ prefix }}--nav-search__template">
+	<div class="{{ base_class }} {{ className|default("") }}" id="{{ base_class }}" tabindex="0">
+		<div class="{{ base_class }}__container">
+			{% include '@components/form/form.twig' with {
+      attributes: {
+        action: search.url,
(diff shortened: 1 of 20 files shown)


packages/twig/src/components/nav/desktop/navsearch.twig
```diff
@@ -0,0 +1,25 @@
+{# navsearch.twig #}
+{% set prefix = prefix|default("ilo") %}
+{% set base_class = prefix ~ "--nav-dropdown" %}
+
+<template id="{{ prefix }}--nav-search__template">
+	<div class="{{ base_class }} {{ className|default("") }}" id="{{ base_class }}" tabindex="0">
+		<div class="{{ base_class }}__container">
+			{% include '@components/form/form.twig' with {
+      attributes: {
```
bashlk, line 9: I think it is better if the variables of the include are indented some more so that the closing bracket `%}` is at the same level as the opening one. 

Ideally these kind of things should be tackled by a linter and lint rules but I'm not sure whether we have one set up for the twig project.

PR #1653, opened 2025-10-07
feat(react): implement loading skeleton components for cards

### LINKED ISSUE #1538: Create skeleton loader for cards

bashlk:
Fixes #1538 

As previously discussed in #1646, this PR re-implements the loading skeletons for the card components as separate components that live alongside the card components. Tests and stories for these new components were also added and the basic test coverage for the card components that was introduced in the previous PR was retained.


packages/react/src/components/Cards/DataCard.tsx
```diff
@@ -188,4 +202,74 @@ const DataCard = forwardRef<HTMLDivElement, DataCardProps>(
...
+            <div className={`${baseClass}--skeleton--label`} />
+            <div className={`${baseClass}--skeleton--copy`} />
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export { DataCard, DataCardSkeleton };
```
justintemps, line 275: @bashlk didn't we agree that the API for the skeletons would be `DataCard.Skeleton`?

justintemps (commented):
Hey @bashlk this mostly looks good, I'm just wondering if the current API is fine or if we should insist on the heuristic we talked about during our call:

Is this better?

```jsx
import { PromoCard } from @ilo-org/react
...
if (loading) {
    return <PromoCard.Skeleton {...props} />
}

return <PromoCard {...props} />
```

Or is this better?

```jsx
import {PromoCard, PromoCardSkeleton } from "@ilo-org/react

if (loading) {
    return <PromoCardSkeleton {...props} />
}

return <PromoCard {...props} />
```

packages/react/src/components/Cards/DataCard.tsx
bashlk: I'm not really a big fan of assigning properties to a functional component to achieve this since it is messing with a JS built-in and leads to the skeleton component being bundled in the output JS whenever a card is used, even if the skeleton is not used.

But if you guys strongly prefer having this structure, I can refactor the skeletons accordingly.

GGKapanadze:
`List` and `Accordion` components already follow the `.` structure as we mentioned during the call. In terms of DX, it’s much better IMHO. The bundle size issue is valid, but I can agree with @justintemps that it will not take up any significant  storage but we can always measure

PR #1679, opened 2025-10-29
React: Photo Gallery component

GGKapanadze:
# Description

This PR introduces a completely new `Photo Gallery` component for the ILO Design system 

https://github.com/user-attachments/assets/737321b5-e2f4-48d8-8228-c015c7de16ea

# Features

## Fit

The design mentions two modes `landscape` and `portrait`, this is translated in the `fit` prop that manipulates how the object fits the container 

| `fit`  | Description | Default |
|--------|--------|--------|
| `contain`  | image is scaled to maintain its aspect ratio while fitting within the element's content |  |
| `cover`  |  image is sized to maintain its aspect ratio while filling the element's entire content box | ✅ |
| `fill`  | image is sized to fill the element's content box |  | 

The idea is to give user the flexibility of css `object-fit`, for future we could allow users to pick the fit by the individual image.

For example `contain` can be used to display the vertical images in a correct shape 
image

## Thumbnails 

An  limited amount of images, that is defined by the width of the carousel that is synced with the controls 

https://github.com/user-attachments/assets/0a7a0a20-00fe-4971-b823-3407074042d7

As described in figma there should be an option to have a two column, this is exposed as a `thubmnailColumns` prop that is defaulted two **1**, here is an example of two column thumbnails 

image

## Controls

Controls are activated during the `hover` and they also have support for the `first` and the `last` item

https://github.com/user-attachments/assets/27be8b8a-0a06-4fd6-a0a5-770f74d56390

## Keyboard Controls 

Keyboard controls are also supported for the "unmaximized" carousel, but they are optional and `off` by default. The reason is not to override the keys by default. The `withKeyboardControls` can be set to `true` to activate the keyboard navigation  

https://github.com/user-attachments/assets/8f7ce7dc-7520-4854-9ac3-b0099cc6c6eb

## Captions

The captions also have the flexibility, there are three modes  of displaying the captions 

| Mode | Behavior | Default |
|--------|--------|--------|
| `visible`  | Caption bar is always shown even if caption is not provided  | ✅ |
| `hidden`  | Caption bar is always hidden |  |
| `ifExists`  | Caption bar operates on demand, if an image has a caption it will be displayed | | 

image

> `visible`

image

> `hidden` 

## The Lightbox

The `Photo Gallery` need a maximization implementation so we need to have a lightbox for it, the activation can be done either by clicking the `Full Screen` | `Zoom`  button or by clicking the `See All` button on thumbnail bar

https://github.com/user-attachments/assets/05275d6d-ea2a-4202-948e-a8c9b3316d9a

Lightbox is completely  synced with the outside carousel, so there are multiple benefits from it

- Choosing an image outside and than clicking the `See All` will open the lightbox with the exact same image
- When closing the lightbox the index will be preserved for the outside carousel

### Thumbnail Carousel

The lightbox implementation of the `thumbnails` is also a separate carousel that is synced with the primary one

https://github.com/user-attachments/assets/a4f66b9d-2323-42c9-897a-fb12280add41

### Controls

The control bar is always visible and `ifExists` strategy  is used for the `captions`. The keyboard navigation including the arrow controls are always enabled

https://github.com/user-attachments/assets/3381a66e-0af4-456c-a3e5-05d0014e4073

### Mobile captions

On mobile, if the caption is longer the viewport it can expand with a light animation

https://github.com/user-attachments/assets/84a77173-e26f-4c53-9a32-92b544236cef

# Implementation Details

Since the component is very complex and has many opinionated implementations it is exposed as one component `<PhotoGallery />` and not multiple `.` ones like `<PhotoGallery.Root>` and `<PhotoGallery.Item`> so api looks like this 

```tsx
<PhotoGallery
  fit="contain"
  items={[
    {
      alt: 'Lorem ipsum dolor sit amet',
      caption: 'Lorem ipsum dolor sit amet consectetur adipiscing elit.',
      credit: 'Photo by John Doe',
      src: '/medium.jpg'
    },
    {
      alt: 'Lorem ipsum dolor sit amet',
      caption: 'Lorem ipsum dolor sit amet consectetur adipiscing elit.',
      credit: 'Photo by Jane Smith',
      src: '/hero.jpg'
    },
  ]}
/>
```
Here is  the full interface of the supported props 

```ts
interface PhotoGalleryProps {
  items: PhotoGalleryItem[];
  fit?: "cover" | "contain" | "fill";
  withKeyboardControls?: boolean;
  thubmnailColumns?: 1 | 2;
  captionView?: "visible" | "hidden" | "ifExists";
}
```

And `PhotoGalleryItem` 

```ts
type PhotoGalleryItem = ImgHTMLAttributes<HTMLImageElement> & {
  credit?: string;
  caption?: string;
  src: string;
  alt: string;
};
```

The component is split into smaller ones, to allow a better maintenance  for the future, plus `Lightbox` and `ExpandableCaption` are abstract enough to be broken as separate ones 

image

## Styles

- The outside carousel is not limited in `width`, we have two height options `420` above `md` and `210` below of it.  If  manipulations are needed, the variable ` --ilo-photo_gallery-target-height` is exposed as mentioned in #999. 
- The thumbnail images  always 16/9 
- The maximized image is always using the native aspect ratio 
- The Image component is used to display the image 

Style files strictly follows the component structure 
image

### DIFF
--- a/packages/react/src/components/PhotoGallery/PhotoGallery.tsx
+++ b/packages/react/src/components/PhotoGallery/PhotoGallery.tsx
@@ -14,16 +14,20 @@ import { LightBox } from "./LightBox";
 import { LightBoxGallery } from "./LightBoxGallery";
 import { Icon } from "../Icon";
 
-export type PhotoGalleryItem = ImgHTMLAttributes<HTMLImageElement> & {
+type PhotoGallerySrc = {
+  gallery: string;
+  thumbnail: string;
+  lightbox: string;
+};
+
+export type PhotoGalleryItem = Omit<
+  ImgHTMLAttributes<HTMLImageElement>,
+  "src"
+> & {
   credit?: string;
   caption?: string;
-  src: string;
+  src: PhotoGallerySrc | string;
   alt: string;
-  sources?: {
-    gallery: string;
-    thumbnail?: string;
-    lightbox?: string;
-  };
 };
 
 export interface PhotoGalleryProps {
@@ -103,7 +107,10 @@ function PhotoGallery({
                     credit={item.credit}
                     url={[
                       {
-                        src: item.sources?.gallery || item.src,
+                        src:
+                          typeof item.src === "string"
+                            ? item.src
+                            : item.src.gallery,
                         breakpoint: 800,
                       },
                     ]}
--- a/packages/react/src/components/PhotoGallery/LightBoxGallery.tsx
+++ b/packages/react/src/components/PhotoGallery/LightBoxGallery.tsx
@@ -86,7 +86,11 @@ function LightBoxGallery({
               <div className={`${coreClass}__slide`} key={index}>
                 <figure className={`${coreClass}__image-wrapper`}>
                   <img
-                    src={item.sources?.lightbox || item.src}
+                    src={
+                      typeof item.src === "string"
+                        ? item.src
+                        : item.src.lightbox
+                    }
                     alt={item.alt || item.credit || `Photo ${index + 1}`}
                   />
                   {item.credit && (
@@ -141,7 +145,11 @@ function LightBoxGallery({
                     onClick={() => onThumbnailClick(index)}
                   >
                     <img
-                      src={item.sources?.thumbnail || item.src}
+                      src={
+                        typeof item.src === "string"
+                          ? item.src
+                          : item.src.thumbnail
+                      }
                       alt={item.alt || item.credit || `Thumbnail ${index + 1}`}
                       loading="lazy"
                     />
--- a/packages/react/src/components/PhotoGallery/PhotoGalleryThumbnails.tsx
+++ b/packages/react/src/components/PhotoGallery/PhotoGalleryThumbnails.tsx
@@ -51,7 +51,9 @@ function PhotoGalleryThumbnails({
                 onClick={() => onSelect?.(index)}
               >
                 <img
-                  src={item.sources?.thumbnail || item.src}
+                  src={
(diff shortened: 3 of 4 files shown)


packages/react/package.json
```diff
@@ -85,6 +85,8 @@
     "@popperjs/core": "^2.11.8",
     "classnames": "^2.3.1",
     "dom-helpers": "^5.2.1",
+    "embla-carousel": "^8.6.0",
```
bashlk: Based on the docs, only the embla-carousel-react library is needed. Currently embla-carousel alone is being used to import a type and I'm sure that it should also be exported from embla-carousel-react

packages/react/src/components/PhotoGallery/PhotoGallery.tsx
```diff
@@ -0,0 +1,147 @@
...
+  items,
+  fit = "cover",
+  withKeyboardControls = false,
+  thubmnailColumns = 1,
+  captionView = "visible",
+}: PhotoGalleryProps) {
+  const { prefix } = useGlobalSettings();
+  const [emblaRef, emblaAPI] = useEmblaCarousel({
+    loop: true,
+    direction: document.dir === "rtl" ? "rtl" : "ltr",
```
bashlk: Can't document.dir be passed directly as the direction?

justintemps:
Hey @GGKapanadze could you have a look at the comments above and get back?

justintemps:
Sorry @GGKapanadze I think may be I didn't click to submit my review, see above 👆

PR #1697, opened 2025-11-18
Photo Gallery

GGKapanadze:
# Description

This PR introduces a `Photo Gallery` component for the twig design system. In terms of features is supports 100% of #1679 and also uses the exact same dependency for the convenience of the maintenance.

# Props

The props  API is the same but it uses the `snake_case` 

| name | default | options |
|--------|--------|--------|
| `see_all_label` | "See All"  | |
| `fit` | "cover" | "cover", "contain", "fill" |
| `with_keyboard_controls` | `false` | `true` | `false` |
| `caption_view` | "visible" | "visible", "hidden", "ifExists" | 

# Directory Structure

The directory structure follows the styles and react package, the only exception is that expandable caption is not a separate component

image

# Mechanics & JS Bundle

The implementation on the twig side differs from the react, since the compositions on the twig side works in a completely different way,  first thing you will notice is that js bundle is kinda **huge**, I tried to design them in a more decoupled way but we need a  cross component reactive way of doing things for twig. Since the photo gallery has 4 tracking carousels in it,  plus controls and the caption logic, I designed the twig component to follow, the main carousel, the idea is that we are tracking only one index, and this rest UI is synced with it.

# Destroy Mechanism

Since the PhotoGallery component is an extremely complex component , there is an addition function called, `destroy` to allow end users to get rid of memory leaks and extra event handlers, because of this everything is the component is designed in a way to be memoized for be destroyed.

### DIFF
--- a/packages/styles/scss/components/photogallery/_photogallery-controls.scss
+++ b/packages/styles/scss/components/photogallery/_photogallery-controls.scss
@@ -15,6 +15,10 @@
     @include typography("highlight-small");
     font-family: var(--ilo-fonts-copy);
     color: var(--ilo-color-white);
+
+    [dir="rtl"] & {
+      flex-direction: row-reverse;
+    }
   }
 
   &__previous-set,
@@ -60,6 +64,12 @@
     &:hover::before {
       @include icon("double_chevron_right", var(--ilo-color-blue));
     }
+
+    [dir="rtl"] & {
+      &::before {
+        transform: translate(-50%, -50%);
+      }
+    }
   }
 
   &__prev {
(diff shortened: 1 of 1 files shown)


GGKapanadze:
@justintemps @inesdgomes  do we support the landscape for the photo gallery ? because the height is too limited and I think a we need a separate design for it

justintemps:
> @justintemps @inesdgomes do we support the landscape for the photo gallery ? because the height is too limited and I think a we need a separate design for it

Hey @GGKapanadze As long as it doesn't break, I don't think we need a separate design.

justintemps (approved):

PR #1796, opened 2026-04-29
Fix SearchField prop types

### LINKED ISSUE #1795: SearchField has invalid prop types
The `SearchField` component's prop types describes a `input` prop that passes through values to the underlying input component directly. However, several values within this `input` prop (e.g. `callback`, `value`) are silently dropped. I suspect that several other props are being silently dropped. Moreover, the types state that `input.value` is a required prop but it is not used.

The prop types should be fixed to describe props that are still being used.

justintemps:
Hey @bashlk, for you to review please. You're right that `<SearchField />` wasn't doing anything with `input.value` so I've now used it to initialize the `searchValue` state. 

However, I also made the `value` prop on `InputProps` optional. 

I'm not sure what you meant by the `callback` prop, that gets called on submit and it works. Use `onInputChange` if you want the callback to fire `onKeyPress`

Fixes #1795

### DIFF
--- a/.changeset/dull-trainers-whisper.md
+++ b/.changeset/dull-trainers-whisper.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/react": patch
+---
+
+**Input:** Make `value` prop optional
--- a/.changeset/olive-beers-mix.md
+++ b/.changeset/olive-beers-mix.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/react": patch
+---
+
+**SearchField:** use `input.value` to initialize search value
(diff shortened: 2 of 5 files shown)


justintemps:
@bashlk actually, it looks like we have two of the same component in the DS: `input.tsx` and `textinput.tsx`. I think may be should just do away with `input.tsx`.

bashlk:
@justintemps I meant the `callback` prop within `input`. As far as I can see, it is still not being used.

From what I see, `SearchField` uses neither `TextInput` nor `Input`. It renders a HTML input element directly.

justintemps:
hey @bashlk yea I know that's why I closed this and opened a new PR: https://github.com/international-labour-organization/designsystem/pull/1797

Let's deprecate the `input` prop and move the rest of the props to the top level.

PR #1797, opened 2026-04-29
React: rationalize SearchField props

### LINKED ISSUE #1795: SearchField has invalid prop types

justintemps:
This rationalizes the props and propTypes for `<SearchField />` while conserving backwards compatibility for consumers who are still using the old props. 

Fixes #1795

### DIFF
--- a/packages/react/src/components/SearchField/SearchField.tsx
+++ b/packages/react/src/components/SearchField/SearchField.tsx
@@ -29,7 +29,6 @@ const SearchField: FC<
       labelSize,
       tooltip,
       theme,
-      value,
       placeholder,
       name,
       id,
@@ -38,7 +37,6 @@ const SearchField: FC<
   ) => {
     // Top-level props win; fall back to the deprecated `input` prop so
     // existing consumers keep working unchanged.
-    const resolvedValue = value ?? input?.value ?? "";
     const resolvedPlaceholder = placeholder ?? input?.placeholder;
     const resolvedName = name ?? input?.name;
     const resolvedLabel = label ?? input?.label ?? "";
@@ -51,7 +49,7 @@ const SearchField: FC<
     // preserve the prior behavior of `input.disabled` (input/button only).
     const inputDisabled = disabled ?? input?.disabled;
 
-    const [searchValue, setSearchValue] = useState<string>(resolvedValue);
+    const [searchValue, setSearchValue] = useState<string>("");
     const { prefix } = useGlobalSettings();
     const baseClass = `${prefix}--searchfield`;
     const buttonClass = `${baseClass}--button`;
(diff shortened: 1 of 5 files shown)


justintemps:
@bashlk I think we should deprecate the `input` prop and move the rest of the props to the top level. Have a look at this and see what you think.

packages/react/src/components/SearchField/SearchField.tsx
```diff
@@ -5,11 +5,53 @@ import { SearchFieldProps } from "./SearchField.props";
...
+      value,
+      placeholder,
+      name,
+      id,
+    },
+    ref
+  ) => {
+    // Top-level props win; fall back to the deprecated `input` prop so
+    // existing consumers keep working unchanged.
+    const resolvedValue = value ?? input?.value ?? "";
```
bashlk: You could do something like this to make it tidier.

```
const inputProps = {
     ...input,
     value,
     placeholder
     (Other values)
}
```

packages/react/src/components/SearchField/SearchField.tsx
```diff
@@ -5,11 +5,53 @@ import { SearchFieldProps } from "./SearchField.props";
...
+      helper,
+      error,
+      errorMessage,
+      disabled,
+      style,
+      labelPlacement,
+      labelSize,
+      tooltip,
+      theme,
+      value,
```
bashlk: This component tries to be both a controlled and an uncontrolled component at once and its quite confusing tbh. 

For example, it exposes this `value` prop which make people think that it is a controlled component, but then changing the value after the initial render doesn't update the value that is being displayed.

Ideally, it should belong to one paradigm. Given that it contains a form, exposes submit events and has its own state, I think it leans more towards being an uncontrolled component. To show this, the `value` and `onInputChange` props should be dropped, leaving only `callback` as the way to interact with the value.

packages/react/src/components/SearchField/SearchField.tsx
```diff
@@ -5,11 +5,53 @@ import { SearchFieldProps } from "./SearchField.props";
 import { Icon } from "../Icon";
 import { FormControl } from "../FormControl";
 
+// @TODO Remove the deprecated `input` prop (and SearchFieldInputProps) in
+// the next major release. Until then, the merge logic below preserves
+// backward compatibility for consumers still passing `input={{ ... }}`.
+
 const SearchField: FC<
```
bashlk, line 12: We should add some tests to this component.

bashlk:
@justintemps I think removing the `input` prop is a good improvement to the component. Left a few other thoughts in the review.

packages/react/src/components/SearchField/SearchField.tsx
justintemps: You can't really do this or all of the input values will get overriden with `undefined`

PR #1799, opened 2026-04-29
React: small versions of Dropdown, Searchfield and TextInput

justintemps:
## Summary

- Adds a new `size` prop (`small` | `large`, defaults to `large`) to **Dropdown**, **SearchField**, and **TextInput** in `@ilo-org/react`. Setting `size="small"` renders the new compact variants. For now, this is React-only — small variants will roll out to the remaining form components in follow-ups.
- Consolidates form-element styles: removes `textinput.scss` in favor of the shared `input.scss`, which now drives styles across all text-based form elements (TextInput, SearchField, Dropdown, NumberPicker, DatePicker, TextArea).
- Removes the unused `<Input />` React component (was never exported, not a breaking change).
- Corrects the small label size on `FormControl` and tidies up `MainNav` and `FormElement` prop wiring.

## Changes

- **React**
  - New `size` prop on `Dropdown`, `SearchField`, `TextInput` (+ args, props, stories)
  - New `SmallFormElements` story showcasing the small variants
  - Removed `Input` component and its files
  - `FormControl` / `FormElement` prop cleanup; `MainNav` props update
- **Styles**
  - Small-size rules added to `_dropdown.scss`, `_searchfield.scss`, `_input.scss`, `_textarea.scss`
  - `_formcontrol.scss` small label size fix
  - Deleted `_textinput.scss` and its export from `index.scss`

## Changesets

- `@ilo-org/react` minor, `@ilo-org/styles` minor — small size for Dropdown/SearchField/TextInput
- `@ilo-org/react` patch — remove unused `<Input />`
- `@ilo-org/react` / `@ilo-org/styles` / `@ilo-org/twig` patch — remove `textinput.scss`


packages/styles/scss/components/_textinput.scss
bashlk, line 1: It seems this removal or another change has caused the styling of the textinput component to break on twig.

Image

packages/react/src/components/SearchField/SearchField.tsx
```diff
@@ -5,11 +5,52 @@ import { SearchFieldProps } from "./SearchField.props";
...
+      placeholder,
+      name,
+      id,
+      size = "large",
+    },
+    ref
+  ) => {
+    // Top-level props win; fall back to the deprecated `input` prop so
+    // existing consumers keep working unchanged.
+    const resolvedPlaceholder = placeholder ?? input?.placeholder;
```
bashlk: This could be made a bit nicer if the variables were grouped into an object.

```
const resolved = {
  placeholder: placeholder ?? input?.placeholder,
  name: name ?? input?.name,
  ...
};
```

packages/react/src/stories/Form/SmallFormElements.stories.tsx
bashlk, line 1: We currently have stories per component and `small` is a variant of the existing components. So this story doesn't really fit this pattern.

justintemps: @bashlk argh this shouldn't have been committed, it was only to show the components side by side

packages/react/src/components/SearchField/SearchField.tsx
justintemps: @bashlk maybe true, but it's throwaway code anyway, it will all be deprecated so who cares?

packages/styles/scss/components/_textinput.scss
justintemps: @bashlk good catch

packages/react/src/components/SearchField/SearchField.tsx
justintemps: @bashlk ok fine I did it, it's pretty, I hope you're happy now.

bashlk: Well it will live in the codebase until its deprecated, which might be in months, so its worth having it a bit nicer I think.

It just bugs me tbh 😆 but it's a very minor thing. I'll leave it to you to decide.

Issue #1627, opened 2025-09-19
CI/CD fails for pull requests from users outside the organization


justintemps:
You can see an example of this [here](https://github.com/international-labour-organization/designsystem/actions/runs/17850470499/job/50780698215?pr=1625).

In the [Test Github Action](.github/workflows/test.yml) we use the `github.repository_owner` to log into the github container registry and start the themes container. 

```yml
      - name: Run theme container
        run: |
          echo "${{ secrets.ILO_BASE_THEME_TOKEN }}" | docker login ghcr.io -u "${{ github.repository_owner }}" --password-stdin
          docker run -d --name theme -p 8082:80 -v "${{ github.workspace }}"/packages/twig/dist/components:/opt/drupal/modules/ilo_base_theme_companion/dist/components -v "${{ github.workspace }}"/packages/twig/dist/styles/index.css:/opt/drupal/modules/ilo_base_theme_companion/dist/css/index.css ghcr.io/international-labour-organization/ilo_base_theme:1.x
```

the problem with this is that, when someone from outside the organization makes a pull request, the tests won't run because their credentials are no good for getting the theme container.

This happens even after I (the design system repository owner) approve running the tests in the Pull Request. 

As we're an open source repository and want to encourage outside contributors to make pull requests, I think we should fix this.

bashlk:
The CI also fails when a PR is made from a branch of a fork of the repository 

<img width="758" height="172" alt="Image" src="https://github.com/user-attachments/assets/d740c6b0-2236-40dc-8309-98e77c881cdb" />

(See: https://github.com/international-labour-organization/designsystem/actions/runs/26326756142/job/77505814381) 

All of these fails seem to stem from the CI not being able to get the base theme image for running the tests, we should probably just make the base theme and its packages public so that they can be pulled without needing any special credentials. The workflow will probably still need to be adjusted to not use any credentials for this step.

PR #1817, opened 2026-05-23
fix(styles): improve dropdown focus styles

### LINKED ISSUE #1806: Incorrect styles for Dropdown component's disabled state
Currently when the Dropdown component is disabled in the dark them, it appears like this

Image

The hover styles also remain active even though the component is disabled

Image

The styles need to be adjusted so that the Dropdown appears disabled visually.

dishanfernando91:
Fixes #1806 

[screen-capture (2).webm](https://github.com/user-attachments/assets/b6ab66a2-bb78-4bff-b24c-20b19b0c43ca)

### DIFF
--- a/packages/styles/scss/components/_dropdown.scss
+++ b/packages/styles/scss/components/_dropdown.scss
@@ -76,6 +76,10 @@
     --ilo-dropdown-color: var(--ilo-color-gray-charcoal);
   }
 
+  &:disabled {
+    pointer-events: none;
+  }
+
   [dir="rtl"] &:hover,
   [dir="rtl"] &:focus,
   [dir="rtl"] &:active {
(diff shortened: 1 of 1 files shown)


justintemps:
I think tests are failing because of #1627, but @dishanfernando91 should be in the organization so I'm not sure what the issue is 🤔 May be that he tried making the PR from a local fork rather than from simply pushing up a new branch to this repo...

dishanfernando91:
> simply pushing up a new branch to this re

I was a bit unclear on the PR process and I did create this from a local fork. I'll go through the contribution docs and check this again.

bashlk:
Yeah https://github.com/international-labour-organization/designsystem/blob/develop/contributing.md does say that PR needs to be made from a fork of a repo and the CI fails when that happens, we should probably fix that. Will add a note to #1627 about this.

Since you are part of the team now @dishanfernando91 you can make a PR directly from the branch of the repo, this is what we usually do.

PR #1822, opened 2026-05-24
fix(styles): improve dropdown focus styles

### LINKED ISSUE #1806: Incorrect styles for Dropdown component's disabled state

dishanfernando91:
Fixes #1806

Dev tested on Firefox v151.x and Chrome v148.x

[dark-mode.webm](https://github.com/user-attachments/assets/51b41801-f781-495c-a5b8-c0accb24d9eb)

### DIFF
--- a/.changeset/pretty-rules-invite.md
+++ b/.changeset/pretty-rules-invite.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": minor
+---
+
+Dropdowns: Update styles to separate enabled and disabled components
--- a/.changeset/pretty-rules-invite.md
+++ b/.changeset/pretty-rules-invite.md
@@ -1,5 +1,0 @@
----
-"@ilo-org/styles": minor
----
-
-Dropdowns: Update styles to separate enabled and disabled components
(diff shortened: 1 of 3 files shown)


bashlk:
@dishanfernando91 Sorry, I think I should elaborated a bit more on the ticket but I think the visuals of the disabled state itself are also not right - see the figma file here https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?node-id=984-0&t=E7MegeIiucEOiWkb-4

We should probably address that as a part of this issue.

dishanfernando91:
> @dishanfernando91 Sorry, I think I should elaborated a bit more on the ticket but I think the visuals of the disabled state itself are also not right - see the figma file here https://www.figma.com/design/RRxsfLNxjoypKrfJg1EwJf/ILO-Components-Library?node-id=984-0&t=E7MegeIiucEOiWkb-4
> 
> We should probably address that as a part of this issue.

Hi @bashlk,

Had to tweak the border opacity. Does need any more adjustments.

image

justintemps (approved):

PR #1823, opened 2026-05-24
fix(styles): update focus handling in buttons

### LINKED ISSUE #1800: Button applies focus styling when clicked
The button component applies focus styling when clicked across all variants. From my memory, it didn't behave like this before.

https://github.com/user-attachments/assets/e156b525-15f7-4a2f-a2ce-9c28e674e167

I believe this is because the focus styling is being applied with the `:focus` selector instead of `:focus-visible`.

dishanfernando91:
Fixes #1800

@bashlk as you suspected focus did persist because of the `:focus` selector. Refactored to use `:focus-visible` instead.

[button_focus.webm](https://github.com/user-attachments/assets/78891666-afc9-449f-8836-abc2efbee7b9)

### DIFF
--- a/packages/styles/scss/components/_button.scss
+++ b/packages/styles/scss/components/_button.scss
@@ -125,13 +125,6 @@
       color: var(--ilo-button-labels-hover-color);
     }
 
-    &:focus {
-      background-color: var(--ilo-color-background-focus);
-      border: var(--ilo-border-md) var(--ilo-color-borders-focus) solid;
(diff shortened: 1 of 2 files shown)


packages/styles/scss/components/_button.scss
```diff
@@ -141,20 +134,23 @@
   }
 
   &:focus {
```
dishanfernando91, line 136: Kept `:focus` to suppress default outline.

bashlk (approved):

PR #1828, opened 2026-05-28
feat(react): implement closed captions

dishanfernando91:
This PR introduced closed captions to the React video player component via a customized `CC` button utilizing underlying the `video.js` `subCapsButton`. Recording for desktop and mobile (andorid) devices attached.

- Added unit tests for the video component
- Created a new story for media with captions.

Chrome 148: 
https://github.com/user-attachments/assets/aa1ea57a-63b9-43d1-9f12-ed2f5b5d5315

Android 16 (tested on Samsung S23 Ultra):
https://github.com/user-attachments/assets/82c1e484-19bf-4afb-97aa-9d451165fd41

### DIFF
--- a/packages/react/src/components/Video/VideoPlayer.props.ts
+++ b/packages/react/src/components/Video/VideoPlayer.props.ts
@@ -4,6 +4,34 @@ import videojs from "video.js";
 export interface VideoPlayerRef {
   player: videojs.Player | undefined;
 }
+
+export interface VideoTextTrack {
+  /**
+   * Source URL for a WebVTT track file.
+   */
+  src: string;
+
+  /**
+   * Track language code (for example "en" or "fr").
+   */
+  srclang: string;
+
+  /**
+   * Human readable language label shown in the CC menu.
+   */
+  label: string;
+
+  /**
+   * Track type rendered by Video.js.
+   */
+  kind?: "captions" | "subtitles";
+
+  /**
+   * Whether this track is selected by default.
+   */
+  default?: boolean;
+}
+
 export interface VideoPlayerControls {
   /**
    * Specify the label for the fullscreen button
@@ -48,4 +76,7 @@ export interface VideoPlayerProps {
    * if YouTube, set to true
    */
   youtube?: boolean;
+
+  // Optional closed-captions tracks
+  tracks?: VideoTextTrack[];
 }
--- a/packages/react/src/components/Video/VideoPlayer.props.ts
+++ b/packages/react/src/components/Video/VideoPlayer.props.ts
@@ -37,12 +37,12 @@ export interface VideoPlayerControls {
   /**
    * Specify the heading shown in the subtitles menu
    */
-  chooseSubtitles?: string;
+  chooseSubtitlesText?: string;
 
   /**
    * Specify the label for the option that turns subtitles off
    */
-  none?: string;
+  noCaptionsText?: string;
 }
 
 export const defaultVideoControls: Required<VideoPlayerControls> = {
@@ -50,8 +50,8 @@ export const defaultVideoControls: Required<VideoPlayerControls> = {
   play: "Play",
   pause: "Pause",
   volume: "Volume",
-  chooseSubtitles: "Choose subtitles",
-  none: "None",
+  chooseSubtitlesText: "Choose subtitles",
+  noCaptionsText: "None",
 };
 
 export interface VideoPlayerProps {
--- a/packages/react/src/components/Video/VideoPlayer.props.ts
+++ b/packages/react/src/components/Video/VideoPlayer.props.ts
@@ -6,29 +6,10 @@ export interface VideoPlayerRef {
 }
 
 export interface VideoTextTrack {
-  /**
-   * Source URL for a WebVTT track file.
-   */
   src: string;
-
-  /**
-   * Track language code (for example "en" or "fr").
-   */
   srclang: string;
-
-  /**
-   * Human readable language label shown in the CC menu.
-   */
   label: string;
-
-  /**
-   * Track type rendered by Video.js.
-   */
   kind?: "captions" | "subtitles";
-
-  /**
-   * Whether this track is selected by default.
-   */
   default?: boolean;
 }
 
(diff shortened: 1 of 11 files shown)


justintemps:
Thanks @dishanfernando91, before I review, it would be good to add a story of a video with captions so @inesdgomes can test.

justintemps:
Also @dishanfernando91 you'll need to add a changeset to this

dishanfernando91:
@justintemps this is just a draft PR I've opened so you or @bashlk could go skim over. There's a few more changes - along with a separate PR for the twig changes - to be made. Once they are done I'll flesh this out and let you guys know.

packages/styles/scss/components/_video.scss
```diff
@@ -210,6 +214,157 @@
...
+        align-items: center;
+        justify-content: center;
+        width: 100%;
+        height: 100%;
+      }
+
+      > .vjs-menu-button .vjs-icon-placeholder:before,
+      > button .vjs-icon-placeholder:before {
+        @include typography("body-xsmall");
+        content: "CC";
```
bashlk: You need to use the closed captions icon from the DS

packages/styles/scss/components/_video.scss
```diff
@@ -210,6 +214,157 @@
...
+        display: none;
+        left: 50%;
+        min-width: px-to-rem(190px);
+        position: absolute;
+        right: auto;
+        transform: translateX(-50%);
+        z-index: 40;
+        padding-top: px-to-rem(28px);
+
+        &:before {
```
bashlk, line 258: Try to find a way to do this without pseudo elements, pseudo elements are not really accessible.

Perhaps you can turn the model into a separate React component whose visibility and positioning is controlled by the VideoPlayer component. You can render it within VideoPlayer.

packages/styles/scss/components/_video.scss
```diff
@@ -210,6 +214,157 @@
...
+          color: var(--ilo-color-gray-charcoal);
+          font-weight: var(--ilo-font-weight-bold);
+          left: px-to-rem(12px);
+          position: absolute;
+          top: px-to-rem(8px);
+        }
+
+        &:after {
+          @include typography("body-xsmall");
+          content: "x";
```
bashlk: Same note about icons.

packages/styles/scss/components/_video.scss
```diff
@@ -210,6 +214,157 @@
...
+        min-width: px-to-rem(190px);
+        position: absolute;
+        right: auto;
+        transform: translateX(-50%);
+        z-index: 40;
+        padding-top: px-to-rem(28px);
+
+        &:before {
+          @include typography("body-xsmall");
+          content: "Choose subtitles";
```
bashlk: The DS doesn't contain its own localizations but the components will be used on multilingual sites so the texts should be customizable.

So add a prop to change this text which defaults to this string.

packages/styles/scss/components/_video.scss
dishanfernando91: Changed

packages/styles/scss/components/_video.scss
dishanfernando91: Changed

packages/styles/scss/components/_video.scss
dishanfernando91: I've added props for both `Choose subtitles` and `None` (to turn off CC)

packages/react/src/components/Video/VideoPlayer.props.ts
```diff
@@ -24,8 +33,27 @@ export interface VideoPlayerControls {
    * Specify the label for the volume button
    */
   volume: string;
+
+  /**
+   * Specify the heading shown in the subtitles menu
+   */
+  chooseSubtitles?: string;
```
bashlk: It would be better if the prop name indicated that this is just a text. i.e. 'chooseSubtitlesText`

packages/react/src/components/Video/VideoPlayer.props.ts
```diff
@@ -24,8 +33,27 @@ export interface VideoPlayerControls {
...
+
+  /**
+   * Specify the label for the option that turns subtitles off
+   */
+  none?: string;
```
bashlk: Suggestion: `noCaptionsText`

packages/react/tests/Video.test.tsx
bashlk, line 1: 👏

dishanfernando91:
@justintemps I might need to revisit some parts of this cause I found a visual bug where the captions button does not stay focused when the track menu is open. 

There are some complexities surrounding how we've approached this component. The control bar being handled entirely with custom CSS makes it difficult to sync the React based track menu with `video.js` captions button. Combine that with handling the mobile user inputs, the logic became a bit brittle to me.

dishanfernando91:
I do also want to know if its possible to make some minor changes to the menu UI. According to the designs, we need to render a `close` button for mobile views and none for desktop.

image

With the implemenation in the PR, we have three ways to perform this. The menu will close when a user clicks/taps outside the menu, presses the captions button again or when a different captions track (or none) is selected. Is it required to show and handle a separate close button as well?

packages/react/src/components/Video/VideoPlayer.props.ts
dishanfernando91: Variable changed

packages/react/src/components/Video/VideoPlayer.props.ts
dishanfernando91: Variable name changed

justintemps:
Hey @dishanfernando91 you're right, we don't need a close button if basically every other interaction you have with the menu closes it. I've seen other video players [like this one](https://cdnapisec.kaltura.com/index.php/extwidget/preview/partner_id/2503451/uiconf_id/43914941/entry_id/1_bbgup9kk/embed/dynamic) and that seems to be the standard behaviour.

dishanfernando91:
@justintemps @bashlk I think I've handled most of the issues I came across with the custom styling. Please take a look at, and if possible have a run at this on Safari.

PR #1833, opened 2026-06-08
feat(twig): handle video captions for drupal/twig components

dishanfernando91:
Provides captioning option for Twig Video Player.

### DIFF
--- a/.changeset/smooth-ladybugs-type.md
+++ b/.changeset/smooth-ladybugs-type.md
@@ -2,4 +2,4 @@
 "@ilo-org/twig": minor
 ---
 
-Video Player: Supports closed captioning
+Video: Supports closed captioning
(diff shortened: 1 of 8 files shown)


.changeset/smooth-ladybugs-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ilo-org/twig": minor
+---
+
+Video player for Drupal components now support closed captioning
```
bashlk: Should follow the same message convention i.e. {component}:

packages/twig/src/components/video/video.component.yml
```diff
@@ -32,7 +32,16 @@ video:
     video:
       type: object
       label: Video
-      description: video player settings
+      description: >-
```
bashlk: Since the structure is described in the fields below, I don't think its necessary to explain it here again.

.changeset/smooth-ladybugs-type.md
dishanfernando91: Message updated.

packages/twig/src/components/video/video.component.yml
dishanfernando91: Reverted

bashlk (approved):

PR #1846, opened 2026-06-13
fix(styles): added ellipsis to long track names in the AudioPlayer

### LINKED ISSUE #1844: AudioPlayer: Long titles shouldn't wrap but should have ellipsis
The title in the audio player should never wrap, it should ellipsis out and stay on one line

Image

bashlk:
Fixes #1844 

This PR adjusts the styling of the AudioPlayer so that long track names don't wrap and instead gets trimmed with ellipsis at the end.

### DIFF
--- a/.changeset/five-houses-sniff.md
+++ b/.changeset/five-houses-sniff.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+AudioPlayer: added ellipsis to long track names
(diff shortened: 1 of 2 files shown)


justintemps:
Thanks for this @bashlk. For future reference, when we make changes to the styles package, I think we should also include the changeset in the twig and react packages so that consumers can see if there are visible changes between one version and the next.

justintemps (approved):

PR #1877, opened 2026-07-30
fix(styles): fixed Profile content alignment when avatar is absent

### LINKED ISSUE #1686: Profile: layout issue when there is no image
When there is no image in the profile component, the text should be fully aligned to the left. 

Image

See here: https://dev-54ta5gq-d6genbv4rqhkw.ch-1.platformsh.site/voices/i-faced-my-hardships-and-it-made-me-stronger

cc @MXTcomunica

bashlk:
Fixes #1686 

This PR fixes the alignment of the content within the Profile component when an avatar is absent. There was an inconsistency in how the Profile was implemented in the React and Twig versions of the DS where one specified the avatar as required and the other specified it as optional so I made it optional in both and ported the fix to both versions of the component.

### DIFF
--- a/.changeset/popular-candles-appear.md
+++ b/.changeset/popular-candles-appear.md
@@ -1,0 +1,7 @@
+---
+"@ilo-org/react": patch
+"@ilo-org/styles": patch
+"@ilo-org/twig": patch
+---
+
+Profile: fixed content alignment when avatar is absent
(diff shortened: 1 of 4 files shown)


GGKapanadze (commented):
Image

Lets add another story with avatar as optional

PR #1879, opened 2026-08-03
Legacy if statement

### LINKED ISSUE #1866: Migrate deprecated Sass `if()` function calls in `_video.scss`
## Description

Since the dependency update that bumped Dart Sass to 1.101.0, every build that compiles our styles (e.g. `@ilo-org/react` `build:docs`) emits deprecation warnings:

​```
Deprecation Warning [if-function]: The Sass if() syntax is deprecated in favor of the modern CSS syntax.

10 │   $box: if($size == "big", 80px, 40px);
   │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    ../styles/scss/components/_video.scss 10:9
​```

Dart Sass has deprecated its `if()` *function* because CSS now has a native `if()` and the two syntaxes conflict. See https://sass-lang.com/d/if-function. The old syntax keeps working until Dart Sass 2.0, so this is warning noise today but will become a hard error eventually.

## Affected code

Exactly two occurrences, both in `packages/styles/scss/components/_video.scss` (lines 9–10):

​```scss
$height: if($size == "big", 72px, 24px);
$box: if($size == "big", 80px, 40px);
​```

## Proposed fix

Do **not** use the modern syntax suggested by the warning (`if(sass($size == "big"): 80px; else: 40px)`) — it only parses on very recent Dart Sass. Since `@ilo-org/styles` ships SCSS source that consumers compile themselves with `sass: ^1.62.1`, the new syntax would turn a warning for us into a hard parse error for downstream users.

Instead, rewrite with a plain `@if` block, which works on every Sass version:

​```scss
$height: 24px;
$box: 40px;

@if $size == "big" {
  $height: 72px;
  $box: 80px;
}
​```

## Acceptance criteria

- [ ] No `if-function` deprecation warnings during builds
- [ ] Video component renders identically at both sizes (big: 80px box / 72px height; default: 40px box / 24px height)
- [ ] Compiled CSS output is unchanged

GGKapanadze:
# Description

This PR removes the deprecated if statement from `scss` implementation

https://github.com/user-attachments/assets/de51b8df-7206-4c19-ada7-2d72f3b9d808

### DIFF
--- a/.changeset/cute-nails-itch.md
+++ b/.changeset/cute-nails-itch.md
@@ -1,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+Video: removed the deprecated if statement
--- a/packages/styles/scss/components/_video.scss
+++ b/packages/styles/scss/components/_video.scss
@@ -6,8 +6,14 @@
 // Video Button
 // ======================================
 @mixin videobutton($name, $size: "standard") {
-  $height: if($size == "big", 72px, 24px);
-  $box: if($size == "big", 80px, 40px);
+  $height: 24px;
+  $box: 40px;
+
+  @if $size == "big" {
+    $height: 72px;
+    $box: 80px;
+  }
+
   $width: $height;
   $icon-color-default: rgba(237, 240, 242, 1);
   $icon-color-hover: rgba(30, 45, 190, 1);


.changeset/cute-nails-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ilo-org/styles": patch
+---
+
+removed the deprecated if statement from the video component
```
bashlk: Usually we prefix the changelogs with the component affected - in this case `Video:`

bashlk (approved):

PR #1903, opened 2026-08-20
Ship all CSS and Sass with react/twig, drop hard dependency on @ilo-org/styles

### LINKED ISSUE #1900: Consumers should only need @ilo-org/react or @ilo-org/twig — no separate @ilo-org/styles install
## Goal

Make `@ilo-org/react` and `@ilo-org/twig` simple to consume. A consumer should be able to:

1. **Install one package** — the component library — and get components *and* stylesheets, without also installing `@ilo-org/styles`.
2. **Include the global stylesheet in one place**, with a single import that just works.
3. **Reach our standard Sass utilities easily** (breakpoints, spacing, typography, mixins) from any of their own `.scss` files, in a way that is guaranteed not to bloat their compiled CSS.

## Current friction

- **Two packages to install and understand.** The component libraries ship compiled CSS but not the Sass source it was built from, so anyone writing their own `.module.scss` must additionally install `@ilo-org/styles`, and our docs have to explain the two-package setup.
- **No safe way to import utilities.** There is no single entrypoint that forwards only functions and mixins. `@use`-ing our existing Sass partials from every module risks duplicating emitted CSS rules into the consumer's bundle, and nothing guards against a partial starting to emit CSS in the future.
- **The global stylesheet import can fail to resolve.** `@ilo-org/react` uses a trailing-slash `exports` entry (`"./styles/"`), deprecated (Node DEP0148) and removed in Node 17, so `@ilo-org/react/styles/index.css` fails to resolve in plain Node. `@ilo-org/twig` has no `exports` map at all.

## Proposed solution

- Add a `scss/_helpers.scss` entrypoint to `@ilo-org/styles` that forwards all functions and mixins from one place and is guaranteed (by a test) to emit no CSS.
- Republish the Sass source under `styles/scss/` in both component libraries, mirroring the `@ilo-org/styles` layout, so the one installed package covers everything:

  ```scss
  @use "@ilo-org/react/styles/scss/helpers" as *;
  ```

- Demote `@ilo-org/styles` from a runtime dependency of the component libraries to a build-time devDependency, so it disappears from consumers' dependency trees.
- Fix the `exports` maps: replace the deprecated trailing-slash export on `@ilo-org/react` with `"./styles/*"` (plus an explicit helpers key), and give `@ilo-org/twig` an equivalent map without breaking existing deep imports.

## Acceptance criteria

- [ ] Installing `@ilo-org/react` (or `@ilo-org/twig`) alone is enough — no `@ilo-org/styles` in the consumer's dependency tree
- [ ] The global stylesheet resolves with one import under Node's `exports` rules (`@ilo-org/react/styles/index.css`)
- [ ] `@use "@ilo-org/react/styles/scss/helpers"` gives access to all standard functions and mixins and emits zero CSS (enforced by a test)
- [ ] Existing imports (compiled CSS paths, `@ilo-org/twig/dist/...` deep imports) keep working

justintemps:
## Summary

Everything a consumer needs — compiled CSS and Sass source — now ships with `@ilo-org/react` and `@ilo-org/twig` themselves, so `@ilo-org/styles` no longer needs to be installed alongside the component libraries. Along the way, splitting the CSS-emitting keyframes out of `_animations.scss` shrinks the compiled CSS bundle by about 5%.

A canary build of these changes is already live in production on [podcast.ilo.org](https://podcast.ilo.org), where it works like a charm.

## Ship all styles with the component libraries (minor: react, twig, styles)

- `@ilo-org/styles` gains a `scss/_helpers.scss` entrypoint that forwards the design system's Sass functions and mixins from a single place. It is guaranteed to emit no CSS (enforced by a test), so it is safe to `@use` from every `.module.scss` without duplicating rules into the compiled bundle.
- `@ilo-org/react` and `@ilo-org/twig` already shipped the compiled CSS under `styles/`; they now also republish the Sass source under `styles/scss/`, mirroring the layout of `@ilo-org/styles`:

  ```scss
  @use "@ilo-org/react/styles/scss/helpers" as *;
  // or, with the Twig:
  @use "@ilo-org/twig/styles/scss/helpers" as *;
  ```

Changes visible to consumers:

- `@ilo-org/styles` moves from a dependency of `@ilo-org/react` and `@ilo-org/twig` to a build-time devDependency, so it no longer lands in consumers' dependency trees.
- `@ilo-org/icons` is now a dependency of `@ilo-org/react`, which the `icon` mixin genuinely needs at compile time (`@ilo-org/twig` already had it).
- The `"./styles/"` export on `@ilo-org/react` is replaced with `"./styles/*"` plus an explicit `"./styles/scss/helpers"` key. Trailing-slash exports are deprecated and were removed in Node 17 — `@ilo-org/react/styles/index.css` previously failed to resolve in plain Node. Existing imports keep working.
- `@ilo-org/twig` gains an exports map with the same `"./styles/*"` and `"./styles/scss/helpers"` keys, plus a `"./*"` passthrough so that existing deep imports like `@ilo-org/twig/dist/components/...` keep resolving.

New `check-exports.mjs` scripts in both packages verify the published exports maps resolve, and the react/twig intro docs are updated to show the new import paths.

## Split the emitting parts out of `_animations.scss` (patch: styles)

`_animations.scss` mixed two pure mixins (`globaltransition`, `pulse-animation`) with five CSS-emitting `@keyframes` blocks, and `_mixins.scss` pulled the whole module in. Every stylesheet that used any mixin therefore inherited all five keyframes, and because legacy `@import` chains do not dedupe, the compiled bundle accumulated hundreds of duplicate copies.

The two mixins now live in a new `scss/_animation-mixins.scss`, which `_mixins.scss` uses instead. `_animations.scss` keeps the keyframes and `@forward`s the mixins, so the 16 component stylesheets that `@use "../animations"` continue to work unchanged.

No API changes and no removed selectors — all non-keyframe CSS is byte-identical. `@keyframes` blocks in the compiled `index.css` drop from 450 to 80, with all five names (`emptygradient`, `pulse`, `slideDown`, `slideUp`, `spin`) intact. Consumers importing `scss/mixins.scss` no longer receive any keyframes at all.

### Bonus: ~5% smaller CSS bundle

Removing the duplicated keyframes reduces the compiled CSS bundle by about 5% (canary build vs. 1.16.0):

| Bundle | 1.16.0 (raw / gzip) | Canary (raw / gzip) | Δ raw | Δ gzip |
| --- | --- | --- | --- | --- |
| `css/index.css` | 582.1 KB / 53.1 KB | 551.3 KB / 52.7 KB | −31.5 KB (−5.3%) | −389 B |
| `css/monorepo.css` | 582.6 KB / 52.8 KB | 551.8 KB / 52.5 KB | −31.5 KB (−5.3%) | −375 B |
| `css/global.css` | 15.9 KB / 3.1 KB | 15.9 KB / 3.1 KB | identical | identical |

## Testing

- A canary release of these packages is deployed to production on [podcast.ilo.org](https://podcast.ilo.org) with no issues.
- New `check-exports.mjs` scripts verify the react and twig exports maps resolve.
- A vitest suite in `@ilo-org/styles` enforces that `scss/_helpers.scss` emits no CSS.

## Other changes

- The canary GitHub Action is updated to work with the new package layout.

Fixes #1900

### DIFF
--- a/.changeset/olive-moons-shine.md
+++ b/.changeset/olive-moons-shine.md
@@ -1,0 +1,43 @@
+---
+"@ilo-org/react": minor
+"@ilo-org/twig": minor
+"@ilo-org/styles": minor
+---
+
+Ship all stylesheets and Sass utilities with `@ilo-org/react` and `@ilo-org/twig`; drop `@ilo-org/styles` as a hard dependency
+
+`@ilo-org/styles` gains a `scss/_helpers.scss` entrypoint that forwards the
+design system's Sass functions and mixins from a single place. It is guaranteed
+to emit no CSS (enforced by a test), so it is safe to `@use` from every
+`.module.scss` without duplicating rules into the compiled bundle.
+
+`@ilo-org/react` and `@ilo-org/twig` already shipped the compiled CSS under
+`styles/`; they now also republish the Sass source under `styles/scss/`,
+mirroring the layout of `@ilo-org/styles`. Everything a consumer needs —
+compiled CSS and Sass — is therefore importable from the component library
+itself, and `@ilo-org/styles` no longer needs to be installed or documented
+alongside it:
+
+```scss
+@use "@ilo-org/react/styles/scss/helpers" as *;
+// or, with the Twig flavor:
+@use "@ilo-org/twig/styles/scss/helpers" as *;
+// or, without a component library:
+@use "@ilo-org/styles/scss/helpers" as *;
+```
+
+Changes visible to consumers:
+
+- `@ilo-org/styles` moves from a dependency of `@ilo-org/react` and
+  `@ilo-org/twig` to a build-time devDependency, so it no longer lands in
+  consumers' dependency trees.
+- `@ilo-org/icons` is now a dependency of `@ilo-org/react`, which the `icon`
+  mixin genuinely needs at compile time (`@ilo-org/twig` already had it).
+- The `"./styles/"` export on `@ilo-org/react` is replaced with `"./styles/*"`
+  plus an explicit `"./styles/scss/helpers"` key. Trailing-slash exports are
+  deprecated (Node DEP0148) and were removed in Node 17 —
+  `@ilo-org/react/styles/index.css` previously failed to resolve in plain Node.
+  Existing imports keep working.
+- `@ilo-org/twig` gains an exports map with the same `"./styles/*"` and
+  `"./styles/scss/helpers"` keys, plus a `"./*"` passthrough so that existing
+  deep imports like `@ilo-org/twig/dist/components/...` keep resolving.
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -62,7 +62,7 @@ jobs:
           pnpm changeset version --snapshot canary-${COMMIT_HASH}
 
           # Publish with canary tag (using changeset publish to respect workspace config)
-          pnpm changeset publish --tag canary --no-git-checks
+          pnpm changeset publish --tag canary --no-git-tag
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -46,13 +46,14 @@
       "import": "./lib/esm/index.js",
       "default": "./lib/cjs/index.js"
     },
-    "./styles/": "./lib/styles/"
+    "./styles/*": "./lib/styles/*",
+    "./styles/scss/helpers": "./lib/styles/scss/_helpers.scss"
   },
   "sideEffects": false,
   "license": "Apache-2.0",
   "types": "./lib/types/index.d.ts",
   "scripts": {
-    "build:lib": "tsc --noEmit --project tsconfig.build.json && vite build",
+    "build:lib": "tsc --noEmit --project tsconfig.build.json && vite build && node scripts/check-exports.mjs",
     "build:docs": "storybook build",
     "check": "tsc --noEmit --p tsconfig.build.json",
     "dev:lib": "vite build --watch",
@@ -79,8 +80,8 @@
   "dependencies": {
     "@ilo-org/brand-assets": "workspace:*",
     "@ilo-org/fonts": "workspace:*",
+    "@ilo-org/icons": "workspace:*",
     "@ilo-org/icons-react": "workspace:*",
-    "@ilo-org/styles": "workspace:*",
     "@popperjs/core": "^2.11.8",
     "classnames": "^2.3.1",
     "dom-helpers": "^5.2.1",
@@ -95,6 +96,7 @@
     "@chromatic-com/storybook": "3.2.7",
     "@ilo-org/eslint-config": "workspace:*",
     "@ilo-org/prettier-config": "workspace:*",
+    "@ilo-org/styles": "workspace:*",
     "@ilo-org/typescript-config": "workspace:*",
     "@storybook/addon-actions": "^8.6.11",
     "@storybook/addon-docs": "^8.6.11",
@@ -124,6 +126,7 @@
     "html-to-image": "^1.11.11",
     "http-server": "^14.1.0",
     "jsdom": "^25.0.0",
+    "sass": "^1.62.1",
     "storybook": "^8.6.11",
     "storybook-addon-rtl": "^1.0.0",
     "vite": "^5.2.0",
--- a/packages/react/scripts/check-exports.mjs
+++ b/packages/react/scripts/check-exports.mjs
@@ -1,0 +1,28 @@
+/**
+ * Asserts this package's exports map still resolves the subpaths consumers are
+ * told to import.
+ *
+ * The Sass surface itself is guarded upstream, by the @ilo-org/styles tests.
+ * What cannot move there is this: the exports map is a property of
+ * THIS package.json, and it is checked with require.resolve rather than a Sass
+ * compile because Sass resolves through loadPaths, which bypasses package
+ * exports entirely and therefore proves nothing about them.
+ *
+ * It runs as part of build:lib rather than as a vitest test because it asserts
+ * against the freshly built lib/, which a bare `vitest` run would not have.
+ */
+import { createRequire } from "node:module";
+
+const require = createRequire(import.meta.url);
+
+for (const subpath of [
+  "@ilo-org/react/styles/index.css",
+  // Node's exports resolution has no directory-index step, so this one resolves
+  // only because of the explicit literal key in the exports map — the
+  // "./styles/*" wildcard would map it to a directory and fail.
+  "@ilo-org/react/styles/scss/helpers",
+]) {
+  require.resolve(subpath); // throws ERR_PACKAGE_PATH_NOT_EXPORTED on regression
+}
+
+console.log("✓ exports resolve");
--- a/packages/styles/scss/_helpers.scss
+++ b/packages/styles/scss/_helpers.scss
@@ -1,0 +1,19 @@
+//*------------------------------------*\
+//    $HELPERS — public Sass API
+//    Functions and mixins only.
+//    Guaranteed to emit no CSS.
+//    Consumers reach this file as `@use "@ilo-org/styles/scss/helpers"` or,
+//    because @ilo-org/react republishes scss/ under styles/scss/, as
+//    `@use "@ilo-org/react/styles/scss/helpers"`.
+//*------------------------------------*/
+
+@forward "functions";
+
+// `dataurlicon` is deliberately withheld from the public API: its default
+// `$color` parameter references `$color-base-neutrals-black`, which is not
+// defined anywhere in this package. Every internal call site passes an explicit
+// colour, so the mixin is dormant there, but a consumer calling it with one
+// argument would hit an undefined-variable error. It is already marked for
+// deprecation in favour of the `icon` mixin. Consumers wanting it can still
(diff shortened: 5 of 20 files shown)


.changeset/olive-moons-shine.md
```diff
@@ -0,0 +1,43 @@
+---
+"@ilo-org/react": minor
+"@ilo-org/twig": minor
+"@ilo-org/styles": minor
+---
+
+Ship all stylesheets and Sass utilities with `@ilo-org/react` and `@ilo-org/twig`; drop `@ilo-org/styles` as a hard dependency
+
+`@ilo-org/styles` gains a `scss/_helpers.scss` entrypoint that forwards the
```
bashlk, line 9: Do we really need to expose the SASS helpers and SASS files from the packages?

IMO these are internals of the DS and exposing them will lead to downstream projects getting coupled to the internal SASS files and breaking whenever they change.

I personally think that the DS package's should only expose styles that would be needed by a consumer to render the component. If we start exposing mixins and others things, we should then consider that a part of the API and document it as well.

.github/workflows/canary.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           pnpm changeset version --snapshot canary-${COMMIT_HASH}
 
           # Publish with canary tag (using changeset publish to respect workspace config)
-          pnpm changeset publish --tag canary --no-git-checks
+          pnpm changeset publish --tag canary --no-git-tag
```
bashlk, line 65: --no-git-tag is not a valid `pnpm publish` argument, this was probably hallucinated. Not sure what the intention behind changing this was.

packages/react/scripts/check-exports.mjs
bashlk, line 1: I don't quite really like this script, things like this should really be handled by a testing framework.

This raises the question of e2e tests. Currently we don't test the consumption / usage of the react package and that includes more things than just this export. So perhaps that's what we should focus on, at least with a single e2e test or two to verify that the package is being built correctly with all its exports.

packages/react/package.json
```diff
@@ -46,13 +46,14 @@
...
-    "./styles/": "./lib/styles/"
+    "./styles/*": "./lib/styles/*",
+    "./styles/scss/helpers": "./lib/styles/scss/_helpers.scss"
   },
   "sideEffects": false,
   "license": "Apache-2.0",
   "types": "./lib/types/index.d.ts",
   "scripts": {
-    "build:lib": "tsc --noEmit --project tsconfig.build.json && vite build",
+    "build:lib": "tsc --noEmit --project tsconfig.build.json && vite build && node scripts/check-exports.mjs",
```
bashlk: If we are leaving this here, it would be better to separate check exports as a separate script.

packages/styles/scss/_helpers.scss
```diff
@@ -0,0 +1,19 @@
...
+//    Functions and mixins only.
+//    Guaranteed to emit no CSS.
+//    Consumers reach this file as `@use "@ilo-org/styles/scss/helpers"` or,
+//    because @ilo-org/react republishes scss/ under styles/scss/, as
+//    `@use "@ilo-org/react/styles/scss/helpers"`.
+//*------------------------------------*/
+
+@forward "functions";
+
+// `dataurlicon` is deliberately withheld from the public API: its default
```
bashlk: Is this comment still relevant given the changes of the other PRs?

packages/styles/tests/helpers.test.mjs
bashlk, line 1: Unrelated note: might be nice to add tests to ensure that only defined tokens are used within the styles package. Although, I guess we also have some user customizable CSS variables in some places 🤔

packages/twig/scripts/check-exports.mjs
bashlk, line 1: A better place for this test might be the base theme. Most consumers are expected to consume the twig DS through that.

packages/twig/package.json
```diff
@@ -45,7 +50,7 @@
     }
   ],
   "scripts": {
-    "build:lib": "rollup -c",
+    "build:lib": "rollup -c && node scripts/check-exports.mjs",
```
bashlk: Same note as react package

.changeset/olive-moons-shine.md
justintemps: @bashlk unfortunately yes, they end up getting used in every design system project that we have. I don't see that changing either, we don't have a layout system and there's no default typography. You're right about having to consider the sass mixins as part of the api, which I put in tests to make sure that they're part of the build. But right now, you practically can't create a full website without reaching into currently undocumented sass internals and that ain't great, even for us!

.github/workflows/canary.yml
justintemps: @bashlk it's not a hallucination, `--no-git-tag` is a [changeset argument](https://changesets.dev/guide/cli#git-tag). Here pnpm is calling the changeset cli.

I guess the argument must have changed, the canary [Github action broke](https://github.com/international-labour-organization/designsystem/actions/runs/32353968297/job/96381535752) with this helpful error message:

```
Run # Get short commit hash for snapshot tag
🦋  All files have been updated. Review them and commit at your leisure
🦋  error Unknown flag for publish: --git-checks
🦋  error Usage: changeset publish [--tag <name>] [--otp <code>] [--no-git-tag]
```

As the error message suggests, changing the command to `pnpm changeset publish --tag canary --no-git-tag` fixed it.

packages/react/scripts/check-exports.mjs
justintemps: @bashlk I had the same intuition at first, but like you say, we don't have e2e tests setup for the react package and honestly, setting those up felt out of scope for this. If we want to add them down the line to ensure the react package is exporting everything it's supposed to, by all means lets do it, but for the time being having something felt better than nothing.

packages/react/package.json
justintemps: @bashlk worth it? we'd still have to run it at build time.

packages/styles/scss/_helpers.scss
justintemps: @bashlk yea you're right I got rid of this.

packages/styles/tests/helpers.test.mjs
justintemps: @bashlk good idea, I created issue #1909

packages/twig/scripts/check-exports.mjs
justintemps: @bashlk we probably need it in both places. Our biggest consumer (ilo.org) doesn't use the base theme.

packages/twig/package.json
justintemps: @bashlk there are so many scripts in each package, are we sure it's worth it? Like, do we really need to test that in isolation? For now I'm happy for it just to break the build.

packages/twig/scripts/check-exports.mjs
```diff
@@ -0,0 +1,31 @@
...
+
+for (const subpath of [
+  "@ilo-org/twig/styles/index.css",
+  // Node's exports resolution has no directory-index step, so this one resolves
+  // only because of the explicit literal key in the exports map — the
+  // "./styles/*" wildcard would map it to a directory and fail.
+  "@ilo-org/twig/styles/scss/helpers",
+  // The "./*" passthrough: this package had no exports map before, so direct
+  // dist/ paths are public API for existing consumers and must keep resolving.
+  "@ilo-org/twig/dist/styles/index.css",
```
GGKapanadze, line 26: we can also check the react package exports here

GGKapanadze (approved):

packages/twig/scripts/check-exports.mjs
justintemps: @GGKapanadze react has its own check-exports.mjs

justintemps: Hey @GGKapanadze react has its own check-exports function: packages/react/scripts/check-exports.mjs

packages/react/package.json
bashlk: I meant just adding it as "check-scripts" here so that each script in package.json has a single function.

packages/twig/package.json
bashlk: I think it would be better if each script did a single thing. That can then be referenced here as `"build:lib":  "rollup -c && pnpm check-exports"`.

packages/react/scripts/check-exports.mjs
bashlk: Temporary things have an annoying habit of becoming permanent. and also mutating to become even uglier things down the line 😆

Personally, I would spin this off as an e2e tests task and handle it well there rather than adding ad-hoc scripts to an already long list of scripts. But I leave it to you.

.changeset/olive-moons-shine.md
bashlk: It might be worth deliberating a bit more about exactly what we expose and take responsibility for instead of exposing everything en-masse and then getting ourselves trapped into a corner later. This would involve some work, of looking at which internals consumers often use and gradually exposing them cleanly from the DS.

IMO the consumers who used the internals of the DS are going down a wrong road and it would be better to nip that in the bud now before allowing it to get worse.

packages/twig/package.json
justintemps: Ok, I've added this

packages/react/package.json
justintemps: done

packages/react/scripts/check-exports.mjs
justintemps: Agreed, but let's do it in a second step. I'll create a ticket after we merge this.

.changeset/olive-moons-shine.md
justintemps: We're heading towards a version 2 release anyway. For the version 1, let's cleanly document everything that's already exposed and use that as a basis to fix the things we don't like in the next major.

PR #1905, opened 2026-08-21
docs(react): overhaul the React Storybook documentation

### LINKED ISSUE #1174: Docs: Include section on Icons
We have a bunch of icons and some components like the button will allow you to include an icon by its name, but we don't have a list of icons anywhere or instructions for how to use them in Twig and React. We need to create a page in both Storybook projects with that information.

### LINKED ISSUE #1269: Update React Docs to include correct usage instructions
The usage instructions for React right now are incorrect. They need to be updated to include the changes introduced in 0.18.0

See below from Changelog.

e631a897f: Introduced a new export system, which allows you to import components, hooks and styles from the library in a more convenient way.

  - `@ilo-org/react/styles` is reserved for the exported styles

  ```jsx
  ...
  import "@ilo-org/react/styles/global.css";
  import "@ilo-org/react/styles/index.css";
  ```

  - `@ilo-org/react` exports all components and hooks

  ```jsx
  import { Loading, Button, useGlobalSettings } from "@ilo-org/react";
  ```

justintemps:
Brings the React package's Storybook docs up to par with the Twig docs, replacing the old minimal pages with a full documentation suite.

Unfortunately no deploy preview for this since it's tacked on top of #1903, so you'll have to pull it down and review locally.

**⚠️ Depends on #1903**

These docs are written against the new consumer-facing API introduced in #1903 (all CSS and Sass shipped with `@ilo-org/react`, no hard dependency on `@ilo-org/styles`): installation instructions, style imports, and the Sass helper entrypoints all assume the new package layout. This PR targets `feat/remove-styles-dep` and should be merged into it **first**, so the docs land in `develop` together with the API change when #1903 merges — but hold off until #1903 itself has been **approved**, since any API changes requested there will need to be reflected here.

### What's new

- **New pages**: Layouts, Theming, Typography, Icons, React Hooks, SASS Helpers, Drupal examples, and Resources
- **New API documented throughout**: Installation, style imports, and Sass usage all reflect the #1903 package layout
- **Icons**: new Usage/Icons page with a live gallery of all 76 icons, usage guidance, and a link to the ILO Icons Figma file; the gallery moved there out of the Icon component story, which now links back to it
- **Rewritten pages**: Introduction, Installation, and Changelog updated and brought in line with the Twig docs
- **Examples**: `examples.mdx` renamed to `react-examples.mdx`, with site screenshots added alongside the new Drupal examples page
- **Storybook chrome**: updated sidebar ordering, shared docs styling (`docs-table.scss` and per-page stylesheets), and manager/preview theming tweaks

No changes to the published library — everything here is documentation and Storybook configuration. `pnpm build:docs` passes.

Fixes #1174 
Fixes #1269


packages/react/src/stories/docs/installation.mdx
GGKapanadze, line 1: I think we should mention the `GlobalProvider`

packages/react/.storybook/manager-head.html
```diff
@@ -18,3 +18,30 @@
...
+      const collapsed = menu.querySelectorAll(
+        'button[aria-expanded="false"][data-nodetype="group"]'
+      );
+      if (collapsed.length === 0) {
+        observer.disconnect();
+        return;
+      }
+      collapsed.forEach((button) => button.click());
+    });
+    observer.observe(document.body, { childList: true, subtree: true });
```
GGKapanadze: this part does nothing for the current state

`[data-nodetype]` lives on the parent element

Image

if we want to expand all nested groups on all levels we should do something like ` [data-nodetype="group"] > button[aria-expanded="false"]` but we don't the auto expanding on twig

packages/react/src/stories/Icon/Icon.stories.tsx
```diff
@@ -61,15 +27,12 @@ const IconDocsPage = () => (
       be rendered in different colors and sizes using props.
     </p>
     <p>
-      Below you can see all the icons currently available in the design system.
+      See the <AnchorMdx href="/docs/usage-icons--docs">Icons page</AnchorMdx>{" "}
```
GGKapanadze: Anchor needs to have `target`

packages/react/.storybook/manager-head.html
justintemps: You're right it does nothing, I just got rid of it @GGKapanadze

packages/react/src/stories/docs/installation.mdx
justintemps: Hey @GGKapanadze I think we should get rid of the GlobalProvider in the next version. It adds nothing and now means that all of our components have to be client components. It already sets the prefix to be "ilo" by default so let's just leave it for now and get rid of it later, what do you think?

justintemps:
Hey @GGKapanadze fixed all of the issues you raised, do you want to give this another look?

packages/react/.storybook/styles.scss
```diff
@@ -9,6 +9,120 @@
...
+    color: var(--ilo-color-brand-800);
+    background-color: var(--ilo-color-brand-100);
+    border-color: var(--ilo-color-brand-200);
+  }
+
+  /**
+    * Code blocks: dark background with Storybook's standard dark-theme
+    * syntax palette.
+    *
+    * Every code surface in the docs — MDX ``` fences and the story
```
bashlk: Nitpick: This comment could be more concise 😆

packages/react/.storybook/styles.scss
```diff
@@ -20,10 +134,62 @@ button[title="Reset zoom"]:hover {
   background-color: var(--ilo-color-brand-200);
 }
 
+// Storybook renders markdown links through its own anchor component, which sets
+// the Storybook UI font (Nunito Sans) on the element itself and ignores <Unstyled>.
+.ilo--richtext a {
```
bashlk: Isn't this redundant given the fix above, which does the same thing?

packages/react/.storybook/styles.scss
```diff
@@ -20,10 +134,62 @@ button[title="Reset zoom"]:hover {
...
   &:hover {
     color: var(--ilo-color-brand-600);
   }
 }
+
+// Design-system components embedded in the docs' richtext wrappers (e.g. the
+// Figma-reference Callouts) pick up richtext's prose styling, which is meant
+// for uncontrolled CMS HTML: h5 gets a large heading margin, p gets the
+// body-large size, and anchors get the full link treatment. Undo those leaks.
+.ilo--richtext {
```
bashlk: I don't think we should be altering DS styles like this because it will also change how the rich text component itself is rendered within the storybook.

This selector should be made more specific so that it only applies to instances of rich text within the doc, not to all richtexts

packages/react/src/stories/docs/styles/welcome.scss
```diff
@@ -1,132 +1,235 @@
...
+    &--copy {
+      display: flex;
+      flex-direction: column;
+      align-items: flex-start;
+    }
 
     &--title {
-      color: rgba(255, 255, 255, 1); // --ilo-color-neutrals-white
+      @include typography("heading-1", true);
+      // Fluid floor: heading-1's 36px mobile size overflows screens narrower
```
bashlk: Nitpick: These `Fluid:` comments are rather redundant.

packages/react/.storybook/styles.scss
justintemps: @bashlk, yea you're right. I'm getting rid of it.

justintemps:
Hey @bashlk thanks for the detailed review, removed most of the extraneous components and fixed the place where you rightly spotted doc styles leaking into stories.

packages/react/src/stories/docs/layouts.mdx
```diff
@@ -0,0 +1,191 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    {[
+      ["xs", "320px"],
+      ["sm", "414px"],
+      ["md", "610px"],
+      ["lg", "1024px"],
+      ["xl", "1140px"],
+      ["xxl", "1168px"],
```
GGKapanadze, line 102: thats a bit inaccurate  for the reacts `useBreakpoint`  we do read the variables but we default `xxl` to 1440 instead of 1168 so we need to adjust it

packages/react/src/stories/docs/typography.mdx
```diff
@@ -0,0 +1,509 @@
...
+    ["--ilo-fonts-monospace", "monospace"],
+  ]}
+/>
+
+To change the system's typefaces, override `--ilo-fonts-display` and `--ilo-fonts-copy` after the stylesheet loads—every component and every mixin output follows along.
+
+### Sizes
+
+Font sizes follow a t-shirt scale. Each token is defined as a `rem` value multiplied by `--ilo-scale` (default `1`), so the whole system can be enlarged or reduced proportionally at runtime by overriding a single variable—`--ilo-scale: 1.125` on a section makes everything in it, type and spacing alike, 12.5% larger. The pixel values below assume the default scale.
+
```
GGKapanadze, line 341: that is factually inaccurate, `--ilo-scale` is defined on `:root` so per section overrides aren't possible

packages/react/src/stories/docs/sass-helpers.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+The examples on this page load it with `as *` so the helpers can be called without a prefix. If you prefer an explicit namespace, drop the `as *` and call them as `helpers.spacing(4)`, `helpers.px-to-rem(...)`, and so on.
+
+Many of the mixins output CSS custom properties (`var(--ilo-...)`), so make sure the design system stylesheet is loaded on the page as described in [Installation](/docs/getting-started-installation--docs).
+
+## Functions
+
+### `spacing($size)`
+
+Returns a length on the design system's spacing scale: `$size` multiplied by the base spacing unit, which is `4px` at the default scale. The result is a `calc()` expression built on the `--ilo-spacing-base` custom property, so spacing responds to changes in `--ilo-scale` at runtime. Non-integer sizes are rounded to the nearest integer.
```
GGKapanadze, line 38: lets also mention the `:root` part here

packages/react/src/stories/docs/styles/theming.scss
```diff
@@ -0,0 +1,42 @@
...
+      border-bottom: 1px solid var(--ilo-color-neutrals-200);
+      vertical-align: middle;
+    }
+
+    code {
+      font-family: monospace;
+      font-size: 0.8125rem;
+      white-space: nowrap;
+    }
+  }
```
GGKapanadze: this styles arelady exists in `docs-table.scss`

packages/react/package.json
```diff
@@ -118,6 +118,7 @@
     "@types/node": "^20.16.0",
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.0",
+    "@types/react-syntax-highlighter": "^15.5.13",
```
GGKapanadze, line 122: lets match the version of `react-syntax-highlighter` and  `@types/react-syntax-highlighter`

packages/react/src/stories/docs/hooks.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+  return breakpoint === "xs" || breakpoint === "sm" ? (
+    <CompactToolbar />
+  ) : (
+    <FullToolbar />
+  );
+};
+```
+
+The hook returns an object with three properties:
```
GGKapanadze, line 41: Image

there is also a fourth one  but we can also mention that its wrapper around the existing library so we can skip the prop declaration at all

GGKapanadze (commented):
This PR introduces a big chunk of documentation that covers the most part of the designsystem, extremely good job 

If we merge this right now we need to maintain two documentation, one for `twig` and another one for `react`, but they share almost all the primitives (the theme part, `style` package). I think we need to create a separate place for those type of docs and keep the theme documentation centralized

packages/react/src/stories/docs/layouts.mdx
justintemps: @GGKapanadze, you're right, it looks like we have a problem:

**packages/styles/scss/theme/_breakpoints.scss**

```scss
$breakpoints-foundation: (
  xs: 320px,
  sm: 414px,
  md: 610px,
  lg: 1024px,
  xl: 1140px,
  xxl: 1168px <-- Is this right?
);
```

**packages/react/src/hooks/useBreakpoint.ts**

```js
const defaults = {
  xs: 320,
  sm: 414,
  md: 610,
  lg: 1024,
  xl: 1140,
  xxl: 1440 <- Or is this right?
};
```

Either we should change the value in `scss` to 1440px (probably better) or the useBreakpoint value to 1168px

packages/react/src/stories/docs/typography.mdx
justintemps: @GGKapanadze it's not inaccurate at all:

```css
:root { --ilo-scale: 1.125; }
.my-section { --ilo-scale: 2; }
```

Everything in `.my-section` will use a scale of 2 instead of 1.125.

packages/react/src/stories/docs/sass-helpers.mdx
justintemps: @GGKapanadze what do you mean?

packages/react/src/stories/docs/styles/theming.scss
justintemps: @GGKapanadze thanks good catch

packages/react/src/stories/docs/hooks.mdx
justintemps: @GGKapanadze we already do mention that it's a wrapper around `use-breakpoint` (which I think is [no longer maintained](https://www.npmjs.com/package/use-breakpoint) by the way, we should maybe consider replacing it). I couldn't find anything in the docs about the fourth return value. If we're not using it anywhere, I'd just leave it. 

But I don't think we can just sent people to the docs of the deprecated library we're using, this seems like the minimal information you need to use the hook.

packages/react/package.json
justintemps: @GGKapanadze It looks like `@types/react-syntax-highlighter` tops out at 15.5.13 on npm and `react-syntax-highligher` doesn't ship its own types.

justintemps:
Hey @GGKapanadze thanks for the careful review, I've gone through everything, please take a moment to give it a look and let me know if it's ok. In terms of extracting the shared parts of documentation so they can be used with the Twig storybook as well, that sounds like a great idea, but let's do it in a separate step. We can create an issue after merging this.

GGKapanadze (approved):

